PR #11, opened 2018-04-19
Conform to the latest Web IDL syntax

saschanaz:
1. [Exposed] extended attribute is now required https://github.com/heycam/webidl/pull/423
2. Nested enum is simply a syntax error
3. `implements` is replaced by interface mixins (https://github.com/heycam/webidl/pull/433), and then [mixins are discouraged if we can write a partial interface](https://heycam.github.io/webidl/#using-mixins-and-partials).


foolip:
@andrenatal can you review this?

foolip:
@saschanaz, how would you like to be credited in https://w3c.github.io/speech-api/webspeechapi.html#acknowledgments?

foolip:
Thanks @saschanaz!

PR #13, opened 2018-06-13
Remove use of "in" in IDL

foolip:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=30011.

Drive-by fix: [SameObject] and whitespace.


saschanaz:
Oh, I was about to open an identical pull request 🤣

Can you also replace `SpeechRecognitionList` into `SpeechRecognitionResultList`? Those are just simple typos.

foolip:
@saschanaz done!

PR #16, opened 2018-06-15
Meta: use HTTPS everywhere

foolip:
Most of these URLs already redirect.


foolip:
This will conflict with other changes, so I'll merge this without review.

Issue #37, opened 2018-08-20
SSML support isn't implemented, remove it from spec?


foolip:
In https://github.com/web-platform-tests/wpt/pull/12568 I found that apparently SSML isn't supported anywhere.

If there is no immediate implementer interest, I suggest removing the SSML bits from the spec.

foolip:
I also asked in https://lists.w3.org/Archives/Public/public-speech-api/2018Aug/0000.html

AdamSobieski:
We could add “ssml” to the [SpeechSynthesisUtterance](https://w3c.github.io/speech-api/#tts-section) interface. When “[text](https://w3c.github.io/speech-api/#utterance-attributes)” is set with a sentence of text, the getter for “ssml” could return SSML wrapping the text content as per: `<speak><p><s>{{text}}</s></p></speak>`. When “ssml” is set with valid SSML content, the getter for “text” could return the innerText of the SSML content.

We could also include a feature support interface. A feature support interface includes methods like hasFeature() or isSupported() [1][2]. This would allow JavaScript introspection of the features supported by an implementation.

An interface like DataTransfer could be utilized to support both “text/plain” and “application/ssml+xml” scenarios [3].

[1] https://www.w3.org/TR/DOM-Level-3-Core/core.html#DOMFeatures
[2] https://wicg.github.io/feature-policy/
[3] http://w3c.github.io/html/editing.html#the-datatransfer-interface

AdamSobieski:
Also would like to indicate an idea from Dominic Mazzoni from the mailing list [1], "perhaps there should be a way for voices to explicitly identify that they support SSML, that way clients would be able to safely include SSML only if the engine will be interpreting it" which suggests looking to the [SpeechSynthesisVoice](https://w3c.github.io/speech-api/#tts-section) interface for indicating support for features.

[1] https://lists.w3.org/Archives/Public/public-speech-api/2018Aug/0004.html

foolip:
Implementation bugs:
* Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=795371
* Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16136734/ (closed)
* Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1425523
* WebKit: https://bugs.webkit.org/show_bug.cgi?id=122208

foolip:
[...]

PR #38, opened 2018-08-27
Remove support for SSML

foolip:
Based on manual testing in https://github.com/web-platform-tests/wpt/pull/12568,
support for SSML has not been implemented in any browser.

Fixes https://github.com/w3c/speech-api/issues/37.


foolip:
This removes the mark event, so tests are up at https://github.com/web-platform-tests/wpt/pull/12689 to match that. https://github.com/web-platform-tests/wpt/pull/12568 could be adapted as well. @saschanaz do you think that would be a good test to have?

saschanaz:
>do you think that would be a good test to have?

Yes, LGTM.

foolip:
This and tests are done now, but as indicated in https://lists.w3.org/Archives/Public/public-speech-api/2018Aug/0001.html, I'll let this sit until the end of the week before merging.

GLRoylance:
**Oppose.**

SSML works on Edge, and I use SSML to speak phoneme strings.

guest271314:
Am not sure what is meant by "Remove support for SSML"? Remove the SSML specification? Or remove mention of SSML at Web Speech API specification? 

One alternative to removing mention of SSML in the Web Speech API specification is to compose code to parse plain text and SSML according to the specification. If there are any inconsistencies in the specification found when composing code reconcile (e.g., different _services_ might currently include a change to prosody or pause before or after `<p>` and `<s>` elements though the specification does not explicitly state a change to  should occur in that context, see https://github.com/guest271314/SpeechSynthesisSSMLParser/blob/master/SpeechSynthesisSSMLParserTest.html#L259; https://github.com/guest271314/SpeechSynthesisSSMLParser/blob/master/SpeechSynthesisSSMLParserTest.html#L274) the specification and the code simultaneously.  Implement the code which parses plain text and SSML at each browser which implements Web Speech API.

Issue #14, opened 2018-06-13
Constructors for speech events


saschanaz:
Currently `new SpeechSynthesisEvent("pending", { utterance: new SpeechSynthesisUtterance() })` works on Edge and Firefox but throws on Chrome.

Edge: only requires the first argument
Firefox: requires both the first and second arguments
Chrome: always throws

Can we spec this to align browser behaviors?

foolip:
When sending https://github.com/w3c/speech-api/pull/45 I didn't read the issue description to spot that Edge and Firefox already disagree about whether the argument should be required.

In this case, I would like to align with Firefox, because otherwise the `utterance` attribute has to be nullable, and the only reason for it would be this constructor. That makes it less clear from looking at the IDL whether it can be null for user-agent-created events.

minorninth:
Interop is good, but what's the use case for constructing one of these?

I suppose it's potentially useful for writing unit tests, but I can't think of anything else.

smaug----:
Testing, and consistency with other events.

PR #45, opened 2018-09-19
Add constructors for all event interfaces

foolip:
Fixes https://github.com/w3c/speech-api/issues/14.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/speech-api/pull/45.html" title="Last updated on Sep 19, 2018, 7:59 AM GMT (01ea11b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/speech-api/45/5f17cce...01ea11b.html" title="Last updated on Sep 19, 2018, 7:59 AM GMT (01ea11b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -185,10 +185,16 @@ enum SpeechRecognitionErrorCode {
     "language-not-supported"
 };
 
-[Exposed=Window]
+[Exposed=Window,
+ Constructor(DOMString type, SpeechRecognitionErrorEventInit eventInitDict)]
 interface SpeechRecognitionErrorEvent : Event {
     readonly attribute SpeechRecognitionErrorCode error;
     readonly attribute DOMString message;
+};
+
+dictionary SpeechRecognitionErrorEventInit : EventInit {
+    required SpeechRecognitionErrorCode error;
+    DOMString message = "";
 };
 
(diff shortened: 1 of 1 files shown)


foolip:
@csharrison I did this when realizing https://chromium-review.googlesource.com/c/chromium/src/+/1231839 would add a new event interfaces without a constructor, so might as well fix this now. Would you mind reviewing?

foolip:
This is now blocked on tests. @csharrison says he might give it a shot.

csharrison:
One more thing: Firefox has implemented SpeechRecognition behind a flag, but none of the event constructors require the init dict either.

foolip:
@csharrison you're right, these are the Gecko IDL files:
https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/SpeechRecognitionEvent.webidl
https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/SpeechRecognitionError.webidl (now renamed to `SpeechRecognitionErrorEvent` in spec)

Since it's behind a flag I hope nobody will mind that in this change I've made all init dicts required so as to avoid nullability and picking arbitrary enum values as the defaults.

Happy to change if objections later follow though.

PR #30, opened 2018-08-06
Add SpeechSynthesisEvent charLength attribute

foolip:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=30004.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/speech-api/pull/30.html" title="Last updated on Aug 6, 2018, 10:45 AM GMT (ca2685e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/speech-api/30/37047fa...ca2685e.html" title="Last updated on Aug 6, 2018, 10:45 AM GMT (ca2685e)">Diff</a>

PR #50, opened 2019-03-07
added charLength to SpeechSynthesisEvent

sjdallst:
There was already a PR about this same concept, but it seems to have become inactive. Creating this PR to get things moving again. Here is the [original discussion](https://www.w3.org/Bugs/Public/show_bug.cgi?id=30004). This change has picked up interest from both Microsoft and Google, so now seems like a good time to make the change.

***
<a href="https://pr-preview.s3.amazonaws.com/sjdallst/speech-api/pull/50.html" title="Last updated on Mar 15, 2019, 5:49 PM UTC (1a78c87)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/speech-api/50/e3b96a8...sjdallst:1a78c87.html" title="Last updated on Mar 15, 2019, 5:49 PM UTC (1a78c87)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -598,6 +598,7 @@ interface SpeechSynthesisUtterance : EventTarget {
 interface SpeechSynthesisEvent : Event {
     readonly attribute SpeechSynthesisUtterance utterance;
     readonly attribute unsigned long charIndex;
+    readonly attribute unsigned long charLength;
     readonly attribute float elapsedTime;
     readonly attribute DOMString name;
 };
@@ -605,6 +606,7 @@ interface SpeechSynthesisEvent : Event {
 dictionary SpeechSynthesisEventInit : EventInit {
     required SpeechSynthesisUtterance utterance;
     unsigned long charIndex = 0;
+    unsigned long charLength = 0;
     float elapsedTime = 0;
     DOMString name = "";
 };
@@ -794,6 +796,11 @@ These events bubble up to SpeechSynthesis.
   <dd>This attribute indicates the zero-based character index into the original utterance string that most closely approximates the current speaking position of the speech engine.
   No guarantee is given as to where charIndex will be with respect to word boundaries (such as at the end of the previous word or the beginning of the next word), only that all text before charIndex has already been spoken, and all text after charIndex has not yet been spoken.
   The user agent must return this value if the speech synthesis engine supports it, otherwise the user agent must return undefined.</dd>
+
+  <dt><dfn attribute for=SpeechSynthesisEvent>charLength</dfn> attribute</dt>
+  <dd>This attribute indicates the length of the text (word or sentence) that will be spoken corresponding to this event.
+  This attribute is the length, in characters, starting from this event's {{charIndex}}.
+  The user agent must return this value if the speech synthesis engine supports it or the user agent can otherwise determine it, otherwise the user agent must return undefined.</dd>
 
   <dt><dfn attribute for=SpeechSynthesisEvent>elapsedTime</dfn> attribute</dt>
   <dd>This attribute indicates the time, in seconds, that this event triggered, relative to when this utterance has begun to be spoken.
--- a/index.bs
+++ b/index.bs
@@ -598,7 +598,7 @@ interface SpeechSynthesisUtterance : EventTarget {
 interface SpeechSynthesisEvent : Event {
     readonly attribute SpeechSynthesisUtterance utterance;
     readonly attribute unsigned long charIndex;
-    readonly attribute unsigned long charLength;
+    readonly attribute unsigned long? charLength;
     readonly attribute float elapsedTime;
     readonly attribute DOMString name;
 };
@@ -606,7 +606,7 @@ interface SpeechSynthesisEvent : Event {
 dictionary SpeechSynthesisEventInit : EventInit {
     required SpeechSynthesisUtterance utterance;
     unsigned long charIndex = 0;
-    unsigned long charLength = 0;
+    unsigned long charLength;
     float elapsedTime = 0;
     DOMString name = "";
 };
@@ -800,7 +800,7 @@ These events bubble up to SpeechSynthesis.
   <dt><dfn attribute for=SpeechSynthesisEvent>charLength</dfn> attribute</dt>
   <dd>This attribute indicates the length of the text (word or sentence) that will be spoken corresponding to this event.
   This attribute is the length, in characters, starting from this event's {{charIndex}}.
-  The user agent must return this value if the speech synthesis engine supports it or the user agent can otherwise determine it, otherwise the user agent must return undefined.</dd>
+  The user agent must return this value if the speech synthesis engine supports it or the user agent can otherwise determine it, otherwise the user agent must return null.</dd>
 
   <dt><dfn attribute for=SpeechSynthesisEvent>elapsedTime</dfn> attribute</dt>
   <dd>This attribute indicates the time, in seconds, that this event triggered, relative to when this utterance has begun to be spoken.
--- a/index.bs
+++ b/index.bs
@@ -606,7 +606,7 @@ interface SpeechSynthesisEvent : Event {
 dictionary SpeechSynthesisEventInit : EventInit {
     required SpeechSynthesisUtterance utterance;
     unsigned long charIndex = 0;
-    unsigned long charLength;
+    unsigned long charLength? = null;
     float elapsedTime = 0;
     DOMString name = "";
 };
--- a/index.bs
+++ b/index.bs
@@ -598,7 +598,7 @@ interface SpeechSynthesisUtterance : EventTarget {
 interface SpeechSynthesisEvent : Event {
     readonly attribute SpeechSynthesisUtterance utterance;
     readonly attribute unsigned long charIndex;
-    readonly attribute unsigned long? charLength;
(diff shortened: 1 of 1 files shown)


sjdallst:
Here's the corresponding test [PR](https://github.com/web-platform-tests/wpt/pull/15732)

travisleithead:
Pinging @foolip, and maybe @rniwa or @marcoscaceres could find an appropriate reviewer from Apple/Mozilla? (Not sure who that might be.)

index.bs
```diff
@@ -795,6 +797,11 @@ These events bubble up to SpeechSynthesis.
   No guarantee is given as to where charIndex will be with respect to word boundaries (such as at the end of the previous word or the beginning of the next word), only that all text before charIndex has already been spoken, and all text after charIndex has not yet been spoken.
   The user agent must return this value if the speech synthesis engine supports it, otherwise the user agent must return undefined.</dd>
 
+  <dt><dfn attribute for=SpeechSynthesisEvent>charLength</dfn> attribute</dt>
+  <dd>This attribute indicates the length of the text (word or sentence) that will be spoken corresponding to this event.
+  This attribute is the length, in characters, starting from this event's {{charIndex}}.
+  The user agent must return this value if the speech synthesis engine supports it or the user agent can otherwise determine it, otherwise the user agent must return undefined.</dd>
```
foolip: The problem of returning undefined mentioned in https://github.com/w3c/speech-api/pull/30#discussion_r207848595 remains to be solved here. If you're also planning to implement this (in Chromium?) we can just do what the implementation will do. There are really only two options:
- Don't expose the `charLength` attribute if it's not supported, so that `'charLength' in SpeechSynthesisEvent.prototype` is false (makes feature detection possible)
- If this is a dynamic and depends on the speech synthesis engine at runtime, make the attribute nullable and say null here instead of undefined. There's some precedence for this in [WebRTC](https://w3c.github.io/webrtc-pc/#rtcerror-interface).

foolip:
@sjdallst @travisleithead thanks for sending this PR! If either of you plan to contribute more to this spec and implementation I'd be happy to give you write access so you can review each others PRs instead of being blocked on me.

index.bs
sjdallst: I agree that nullable is the way to go, since it ends up being dynamic based on which speech synthesis engine you are using. Should this logic apply to charIndex, elapsedTime, and name as well? The current implementations seem to return the default values for these fields instead of undefined, and it might cause some compat issues if this is changed.

foolip: Do you mean that it returns 0? If making the distinction between zero and unknown isn't important to clients of the API we could change the spec to do the same. Bull null is fine too.

Would you be up to sending a PR to change the other attributes to change in a similar way?

index.bs
```diff
@@ -598,13 +598,15 @@ interface SpeechSynthesisUtterance : EventTarget {
...
     readonly attribute unsigned long charIndex;
+    readonly attribute unsigned long? charLength;
     readonly attribute float elapsedTime;
     readonly attribute DOMString name;
 };
 
 dictionary SpeechSynthesisEventInit : EventInit {
     required SpeechSynthesisUtterance utterance;
     unsigned long charIndex = 0;
+    unsigned long charLength;
```
foolip:
```suggestion
    unsigned long charLength? = null;
```

marcoscaceres:
@andrenatal from the Mozilla side, perhaps?

index.bs
```diff
@@ -598,13 +598,15 @@ interface SpeechSynthesisUtterance : EventTarget {
 interface SpeechSynthesisEvent : Event {
     readonly attribute SpeechSynthesisUtterance utterance;
     readonly attribute unsigned long charIndex;
+    readonly attribute unsigned long? charLength;
```
marcoscaceres: why null instead of just zero?

foolip: This is related to the "return undefined" bits in https://w3c.github.io/speech-api/#speechsynthesisevent-attributes which don't actually make sense, it's not possible to return undefined through Web IDL. Using `elapsedTime` zero, `name` empty string would be fine, and `charLength` zero too. It's only for `charIndex` that zero may be ambiguous with a real value.

However, with `charLength` added, it would be possible to distinguish this case by checking if `charLength` is zero. @sjdallst do you think that would be OK? The existing Chromium implementation just returns the default values, right?

marcoscaceres: Argh, the above doesn’t sound amazing if it’s going to require special IDL treatment. Maybe we need a more in-depth spec review?

foolip: It won't be hard to find other things that aren't quite right or not precise enough, but this "return undefined" problem is quite confined, it doesn't interact with other problems in the spec AFAICT.

foolip:
@sjdallst, what are your thoughts on the overall "return undefined" situation? Do you think it's OK to just return the default values instead, or should we go with null everywhere?

index.bs
sjdallst: I think that the current state of things is a little ambiguous. If the user agent neglects charIndex completely, the client will see "undefined". If I get a 0 for charIndex, is this because the word or sentence is at the beginning of the text, or is it because the speech synthesizer does not support the charIndex? Returning null would remove some of this ambiguity, so a check for undefined or null would unambiguously tell the client whether this feature is supported, thus allowing for feature checking. Assuming the user agent follows the spec, this would also tell the client who is at fault (user agent or speech synthesis engine)

The current norm is the ambiguous option. I don't know if clients actually code for this ambiguous case, do platform checking, or hope for the best and let themselves break when features aren't supported, but following this norm would probably allow for the easiest and quickest implementation across browsers, and would let us resolve the undefined issue. However, issues might come later when clients have to keep track of what combination of browsers and platforms support these features.

foolip: Are you planning to implement this in Chromium, and have you had a chance to check what the current implementation does yet?

GurpreetV: I have reviewed both the iterations and I am also inclined to setting charLength = 0 because it seems more consistent with how charIndex = 0 and name=emptyString defaults are being used. And since 0 is an organic valid value of an unsigned long, it will be more smooth to implement this in the browser (passthrough between layers) than a null value. Additionally charLength=0 is not a valid value and so will not be ambiguous.

marcoscaceres: > If I get a 0 for charIndex, is this because the word or sentence is at the beginning of the text, or is it because the speech synthesizer does not support the charIndex?...  so a check for undefined or null would unambiguously tell the client whether this feature is supported

I'm not sure I understand: If the user agent doesn't support this feature, then why would the attribute be there? Wouldn't checking the prototype chain (I.e.,  `"charLength" in SpeechSynthesisEvent.prototype` - or variations on that) be the right way to do this? That would remove the ambiguity without needing to have a `length` with two types.

sjdallst: The problem arises when the user agent (chromium, etc) supports the feature, but the speech synthesizer does not. For example, Windows, Mac, and, I think, Android have clear implementations for character length from speech events, but Linux (as far as I know) does not. Therefore, it makes sense to implement this feature in Chromium because it is supported by many speech synthesis platforms, but because the Linux version of chromium cannot get the value, it will be stuck returning a default value. 

In order to do the prototype chain checking, I think a web platform like chromium would have to keep different versions of the SpeechSynthesisEvent for different platforms depending on whether or not that platform's speech synthesizer supports the attribute. This sounds like it would add a lot of unecessary code for any browser implementer that wanted to do this.

I think that if we just look at charLength, it's not a bad idea to return the default value. Word and sentence events are only fired when a word or sentence is about to spoken, and there aren't any speakable words of 0 characters. 

Maybe charIndex is something that should be looked at in another PR?

marcoscaceres: > Linux version of chromium cannot get the value, it will be stuck returning a default value.

What's preventing this getting fixed on Linux tho? It seems like we are working around an OS limitation that could be fixed at the OS level at any time in the future.

marcoscaceres: (also noting that `null` adds further fingerprinting surface because it reveals something unique about the OS's capabilities - or lack of... not a huge thing, but it's a thing additional to consider)

foolip:
@sjdallst can you summarize what browsers currently do on different platforms if the API is supported but the synthesizer turns out not to at runtime?

index.bs
foolip: While I'm leaning towards not nullable because I suspect it aligns with existing implementations, note that if we did make it nullable and then all OS limitations were resolved, it would just mean that null is never returned and we can eventually make it non-nullable again.

sjdallst: Ok it seems to me that we're leaning towards not nullable. I'll go ahead and make the change; along with edits to the descriptions of the other attributes.

sjdallst:
> @sjdallst can you summarize what browsers currently do on different platforms if the API is supported but the synthesizer turns out not to at runtime?

Yeah, right now default values from SpeechSynthesisEventInit are being used, (so 0 for charIndex, etc)

sjdallst:
> @sjdallst, what are your thoughts on the overall "return undefined" situation? Do you think it's OK to just return the default values instead, or should we go with null everywhere?

I think the discussion here has led us towards doing default values. Nullable goes against current practices, and would eventually be phased out once a browser supports the full SpeechSynthesisEvent on all platforms.

foolip:
Thanks for working through this and fixing the other attributes, @sjdallst!

PR #32, opened 2018-08-06
Remove the SpeechGrammar constructor

foolip:
Nothing can be done with these objects, as SpeechGrammarList is
populated using addFromURI/addFromString.

Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=26992.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/speech-api/pull/32.html" title="Last updated on Aug 6, 2018, 10:58 AM GMT (0d9f88d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/speech-api/32/37047fa...0d9f88d.html" title="Last updated on Aug 6, 2018, 10:58 AM GMT (0d9f88d)">Diff</a>

PR #58, opened 2019-08-02
BREAKING CHANGE: Drop grammars

### LINKED ISSUE #57: grammar example
You should provide an fully working example about how to use grammar. Because I did not see any use case where I could use that.

I do not understand what is the purpose of having grammar and how to use it.
"google search" did not help me.

marcoscaceres:
grammars are underspecified - they lack a format, so they are kinda useless... so removes grammars attribute, drops SpeechGrammarList, and SpeechGrammar.

Closes #32
Closes #57

***
<a href="https://pr-preview.s3.amazonaws.com/marcoscaceres/speech-api/pull/58.html" title="Last updated on Mar 5, 2021, 9:38 PM UTC (c51dc49)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/speech-api/58/30e8a49...marcoscaceres:c51dc49.html" title="Last updated on Mar 5, 2021, 9:38 PM UTC (c51dc49)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -69,9 +69,7 @@ This subset does not preclude future standardization of additions to the markup,
 <ul>
   <li>Voice Web Search</li>
   <li>Speech Command Interface</li>
-  <li>Domain Specific Grammars Contingent on Earlier Inputs</li>
   <li>Continuous Recognition of Open Dialog</li>
-  <li>Domain Specific Grammars Filling Multiple Input Fields</li>
   <li>Speech UI present when no visible UI need be present</li>
   <li>Voice Activity Detection</li>
   <li>Temporal Structure of Synthesis to Provide Visual Feedback</li>
@@ -91,8 +89,6 @@ This does not preclude adding support for this as a future API enhancement, and 
 <ul>
   <li>Rerecognition</li>
 </ul>
-
-<p>Note that for many usages and implementations, it is possible to avoid the need for Rerecognition by using a larger grammar, or by combining multiple grammars &mdash; both of these techniques are supported in this specification.</p>
 
 <h2 id=security>Security and privacy considerations</h2>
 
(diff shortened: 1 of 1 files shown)


saschanaz:
What's the implementation status on Chrome? What does it support?

foolip:
I'd like to hold off on removing this from the spec until I understand a bit better why this was added and if it ever worked, or perhaps still works on some platform.

kdavis-mozilla:
@foolip I'd speculate that when the spec was initially written, many years ago now, the production STT engines were not able to handle LVCS and required grammars to get reasonable results. Things have changed somewhat now.

marcoscaceres:
Noting also that `src` attribute has implied fetching behavior and semantics, yet the spec doesn't actually specify how things are fetched, parsed, error handling (e.g., what happens if one fails, the rest download, is order significant, and so on), referrer policy, etc., etc. so, there is no way anyone implemented any of this in an interoperable manner. 

Although well intentioned, fetching such grammars should really be done with `fetch()` or even god ol' xhr if dealing with XML.

foolip:
@marcoscaceres there's no argument that the spec defines this well enough for interoperable implementation, and doing nothing wouldn't resolve https://github.com/w3c/speech-api/issues/57. I've poked some more in https://github.com/w3c/speech-api/issues/57#issuecomment-518126285 and my only concern with dropping this is what then happens with the implementation in Chromium, since I can't guess yet what the impact of removing it would be. Would you be OK with waiting for usage numbers before we axe it from the spec, to avoid flip-flopping? I'd of course not expect anyone to try implementing it in the interim, and warnings to that effect would be fine.

marcoscaceres:
Yes, absolutely: no objection and no rush :)

maccman:
Any update? Does Chrome support SpeechGrammarList?

evanbliu:
The WebSpeech API implementation in Chrome supports SpeechGrammarList. However, the backend speech recognition service does not so they don't actually do anything. I'm in favor of dropping this from the spec.

beaufortfrancois:
With https://github.com/WebAudio/web-speech-api/pull/117 now being merged, I believe we can close this PR.

PR #68, opened 2019-10-07
chore: remove redudant files

marcoscaceres:
Removes the old and duplicate spec and the unmaintained errata file.


foolip:
There are URLs that redirect to these, so they need to stay. Two of them already say "This document and the errata are no longer maintained. For the latest version, please see https://w3c.github.io/speech-api/." Adding that to speechapi-errata.html as well would be good.

marcoscaceres:
Ok, they are now part of https://github.com/w3c/w3c.github.io/pull/72 :)

foolip:
The URLs that should work are:
https://w3c.github.io/speech-api/ (redirect, works)
https://w3c.github.io/
https://w3c.github.io/

foolip:
The URLs that should work are:
https://w3c.github.io/speech-api/ (redirect, works)
https://w3c.github.io/speech-api/speechapi-errata.html (works)
https://w3c.github.io/speech-api/speechapi.html (redirect, broken)
https://w3c.github.io/speech-api/webspeechapi.html (redirect, broken)

I'll send a fix for the last two.

foolip:
I've sent https://github.com/w3c/w3c.github.io/pull/89, but will merge this because those broken redirects aren't critical, and in any case the existence of the files in this repo doesn't help in any way.

PR #73, opened 2019-11-27
Meta: create w3c.json

foolip:


w3c.json
```diff
@@ -0,0 +1,5 @@
+{
+    "group": ["80485"],
+    "contacts": ["marcoscaceres"],
```
foolip: @marcoscaceres given that this is your name here, can you review too?

PR #77, opened 2020-06-25
chore: Use Python 3 for Travis, not Python 2

sideshowbarker:
Bikeshed now requires Python 3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ branches:
     - master
 language: python
 python:
-  - "2.7"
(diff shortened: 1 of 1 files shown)


marcoscaceres:
slightly boring aside, but there might be quite a few specs still using ol' python...
https://github.com/search?q=org%3Aw3c+pip+install+--editable+bikeshed&type=Code

Additionally, seems like specs are wasting a ton of energy installing BikeShed from scratch, when they could just generate directly by using `curl`.

PR #122, opened 2024-12-23
Add on-device speech recognition support to the Web Speech API

### LINKED ISSUE #108: Offline/on-device speech recognition
There is privacy concern that audio data captured in SpeechRecognition API is sent to third-party servers for processing (a similar issue in SpeechSynthesis API: https://github.com/WICG/speech-api/issues/98). Actually not all browsers will send the data to third-party servers, but there is no way for developer to tell with existing API. Developers may check the implementation of different browsers and enable the feature on browsers with offline support, but this solution is not sustainable as browsers may add support for offline mode later, and browsers may switch between modes based on other conditions (networking, language, etc).

To address this concern, we should add a new interface to indicate online/offline mode for recognition. We might also let developer set the mode, as browsers may support different modes.

evanbliu:
Closes #108

This PR updates the Web Speech API spec to include support for on-device speech recognition.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,8 +1,9 @@
 <pre class='metadata'>
 Title: Web Speech API
 Level:
-Status: CG-DRAFT
-Group: Audio Working Group
+Status: WD
+Group: audiowg
+TR:
 URL: https://webaudio.github.io/web-speech-api/
 Repository: WebAudio/web-speech-api
 Shortname: speech-api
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1,8 +1,9 @@
 <pre class='metadata'>
 Title: Web Speech API
 Level:
-Status: CG-DRAFT
-Group: Audio Working Group
+Status: WD
+Group: audiowg
```
hoch, line 5: I think we can leave these out from the PR.

Technically the Web Speech API is still managed by Audio CG, not WG.

hoch (approved):
LGTM with nits

KuayMongkolchai7726 (commented):
Thank Your

index.bs
```diff
@@ -155,12 +157,15 @@ interface SpeechRecognition : EventTarget {
...
     attribute boolean interimResults;
     attribute unsigned long maxAlternatives;
+    attribute SpeechRecognitionMode mode;
 
     // methods to drive the speech interaction
     undefined start();
     undefined start(MediaStreamTrack audioTrack);
     undefined stop();
     undefined abort();
+    boolean onDeviceWebSpeechAvailable(DOMString lang);
```
Rob--W, line 167: At a first glance, this property looks like an event due to the "on" prefix.

A way to avoid that misinterpretation would be to prepend another prefix to the name, e.g. `isOnDeviceWebSpeechAvailable`.

P.S. This is random feedback that does not carry any weight. I just read this proposal out of interest and noticed this.

PR #123, opened 2024-12-23
Update the Web Speech API spec to be a working draft

evanbliu:
This PR updates the Web Speech API spec to be a working draft under the Audio Working Group.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,8 +1,8 @@
 <pre class='metadata'>
 Title: Web Speech API
 Level:
-Status: WD
-Group: audiowg
+Status: CG-DRAFT
+Group: audiocg
 TR:
 URL: https://webaudio.github.io/web-speech-api/
 Repository: WebAudio/web-speech-api
--- a/index.bs
+++ b/index.bs
@@ -1,8 +1,9 @@
 <pre class='metadata'>
 Title: Web Speech API
 Level:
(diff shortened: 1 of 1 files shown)


hoch:
@svgeesus Chris - The current WG charter doesn't cover the work of the Web Speech API yet, so my understanding is that this technical report cannot be published as a WD under Audio WG.

If that's the case, what do we use for `Status` and `Group` field?

svgeesus:
> If that's the case, what do we use for `Status` and `Group` field?

`CG-DRAFT` and `audiocg`

evanbliu:
I made the change, but the CSS spec preprocessor says that "audiocg" isn't a valid group. Should I create a new group somewhere?

padenot:
I think it is `audio-comgp` as in https://www.w3.org/community/audio-comgp/.

dontcallmedom:
IIUC it needs to be added to bikeshed to be recognized - I think in https://github.com/speced/bikeshed-boilerplate/blob/e3237002c87440fbb5cbf7ebed15fafb65dd953d/boilerplate/doctypes.kdl#L104

evanbliu:
Sounds good, I've sent out this PR to add audiocg: https://github.com/speced/bikeshed-boilerplate/pull/114

svgeesus (approved):

PR #132, opened 2025-01-21
Minor spec updates

evanbliu:
This PR contains the following minor spec updates:

- Fix a closing tag to match the opening tag.
- Explicitly mention the risk of fingerprinting in the Security considerations section.
- Update the onDeviceWebSpeechAvailable() and installOnDeviceSpeechRecognition() methods to return a Promise that resolves to a boolean. These will be asynchronous operations in most cases.

***
<a href="https://pr-preview.s3.amazonaws.com/evanbliu/speech-api/pull/132.html" title="Last updated on Feb 14, 2025, 11:04 PM UTC (3c984a1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/132/9c092a6...evanbliu:3c984a1.html" title="Last updated on Feb 14, 2025, 11:04 PM UTC (3c984a1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -165,8 +165,8 @@ interface SpeechRecognition : EventTarget {
     undefined start(MediaStreamTrack audioTrack);
     undefined stop();
     undefined abort();
-    Promise<boolean> onDeviceWebSpeechAvailable(DOMString lang);
-    Promise<boolean> installOnDeviceSpeechRecognition(DOMString lang);
+    static Promise<boolean> availableOnDevice(DOMString lang);
+    static Promise<boolean> installOnDevice(DOMString lang);
 
     // event methods
     attribute EventHandler onaudiostart;
--- a/index.bs
+++ b/index.bs
@@ -238,6 +238,7 @@ dictionary SpeechRecognitionEventInit : EventInit {
     unsigned long resultIndex = 0;
     required SpeechRecognitionResultList results;
 };
+</pre>
 
 <h4 id="speechreco-attributes">SpeechRecognition Attributes</h4>
 
--- a/index.bs
+++ b/index.bs
@@ -120,7 +120,7 @@ This does not preclude adding support for this as a future API enhancement, and 
 
   <li>The user agent may also give the user a longer explanation the first time speech input is used, to let the user know what it is and how they can tune their privacy settings to disable speech recording if required.</li>
 
-  <li>User agents must obtain explicit and informed user consent before installing on-device speech recognition languages that differ from the user's preferred language or when the user is not connected to an Ethernet or Wi-Fi network.</li>
+  <li>To reduce the risk of fingerprinting, user agents must obtain explicit and informed user consent before installing on-device speech recognition languages that differ from the user's preferred language or when the user is not connected to an Ethernet or Wi-Fi network.</li>
 </ol>
 
 <h3 id="implementation-considerations">Implementation considerations</h3>
@@ -147,7 +147,7 @@ This does not preclude adding support for this as a future API enhancement, and 
 The term "final result" indicates a SpeechRecognitionResult in which the final attribute is true.
 The term "interim result" indicates a SpeechRecognitionResult in which the final attribute is false.
 
-<pre class="idl">
+<xmp class="idl">
 [Exposed=Window]
 interface SpeechRecognition : EventTarget {
     constructor();
@@ -165,8 +165,8 @@ interface SpeechRecognition : EventTarget {
     undefined start(MediaStreamTrack audioTrack);
     undefined stop();
     undefined abort();
-    boolean onDeviceWebSpeechAvailable(DOMString lang);
-    boolean installOnDeviceSpeechRecognition(DOMString lang);
+    Promise<boolean> onDeviceWebSpeechAvailable(DOMString lang);
+    Promise<boolean> installOnDeviceSpeechRecognition(DOMString lang);
 
     // event methods
     attribute EventHandler onaudiostart;
@@ -264,7 +264,7 @@ interface SpeechGrammarList {
     undefined addFromString(DOMString string,
                     optional float weight = 1.0);
 };
-</pre>
+</xmp>
 
 
 <h4 id="speechreco-attributes">SpeechRecognition Attributes</h4>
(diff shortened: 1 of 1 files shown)


evanbliu:
@hoch - Can you please take a quick look at this when you get a chance? Let me know if you'd like me to assign this to a different reviewer :)

index.bs
```diff
@@ -165,8 +165,8 @@ interface SpeechRecognition : EventTarget {
     undefined start(MediaStreamTrack audioTrack);
     undefined stop();
     undefined abort();
-    boolean onDeviceWebSpeechAvailable(DOMString lang);
-    boolean installOnDeviceSpeechRecognition(DOMString lang);
+    Promise<boolean> onDeviceWebSpeechAvailable(DOMString lang);
+    Promise<boolean> installOnDeviceSpeechRecognition(DOMString lang);
```
padenot, line 169: This might want a size.

hoch: I am curious - what do you mean by a `size`?

evanbliu: I'm also curious what you mean by a size. Do you mean the size of the on-device speech recognition models? I think those details should abstracted away from the website--browsers can choose to expose that information to users if/when prompting users to download language packs. Otherwise the API would expose an unnecessary bit of information that could be used for fingerprinting.

padenot: Right, so this needs to be talked about, and might hint at the fact that it is premature to do API design, and we should go back to system design.

Text-to-speech models differ greatly in size and capabilities. The model size can influence what languages can be recognized, what accent can be recognized, and generally the accuracy. It will also influence the speed at which the recognition can happen, potentially. There are models that are generic, and models that are more specialized. For a given language, there can be multiple suitable models, and they will not perform the same: a small model trained on some UK or some American accent can be fairly small, while a model that can recognize English with a strong French accent might have to be bigger.

Because one of the goal of this API is to perform real-time on-device recognition, we need to be cognizant of the fact that:

- a language identifier might not enough
- there might be more than one model per language
- those models have widely different inference performance characteristics, both in speed, memory footprint and accuracy
- they also are widely different in size, that in turns influences performance
- bigger, more generic models might not work in real-time on some devices, or at all
- some models are specific to a language, and some to a regional variant of a language

All of this ties heavily into the feature detection discussion we're having in parallel.

Performance **is** a feature here in the sense that it makes it or breaks the feature for the majority of the user base (non-native English speakers), per my (ongoing) survey of the available models.

hoch: Thanks for laying out useful topics.

Let's add this to our teleconference agenda.

evanbliu: All valid points. On Chrome for example, we have multiple models available for the same locale that vary in terms of size, accuracy, and resource consumption. In addition, we have multiple on-device speech recognition models for different locales of the same language (i.e. en-AU and en-GB) as well as on-device LLMs that could potentially support multiple languages and locales. There are also research teams working on personalized speech recognition models specifically trained on an individual, which would dramatically increase accuracy for users with speech impediments. However, I don't believe these implementation details need to be exposed to websites. As you mentioned, there are a ton of factors that go into whether on-device speech recognition works. Exposing all of these factors would open up a significant fingerprinting surface and most websites wouldn't need to know about these details anyways. Ideally, websites can just ask "can this user run on-device speech recognition in X language effectively?" and the browser will do everything else to ensure an optimal speech recognition experience.

padenot: Thanks a lot for your comprehensive response. There are prior art for solving this kind of problem, most notably in WebRTC land. This involved UA-controlled UX elements, that aren't exposed to the page, but that the user can use to inform what the API will do, without revealing anything to the page.

In essence, instead of having a JavaScript API that has quite a lot of parameters to select a model, the API is async, and then this display a browser-controlled piece of UI, and then the user decides something, the promise resolves, and all is good.

As an illustration, let me describe two APIs that allow the same capability, but the seconde one is more privacy preserving:

- `navigator.mediaDevices.enumerateDevices` can enumerate the input and output media devices on the computer into an array, and then the website can display a dropdown with the device name and characteristics. The user picks the one that they want to use. After the selection, the `id` for this device is used, to, e.g. route the audio to the right output device. This exposes a number of bits of entropy that is extremely high, allowing the user to be uniquely identified.
- The alternative API, `selectAudioOuput()` causes a UA-controlled UI element to be displayed, when called by the page. The user chooses the device to use, and validates the choice. This resolves the promise, the page can then route the audio output to this device using its `id` in very much the same way. This only exposes a single device to the page, and is extremely easy to implement by authors.

evanbliu: On Chrome, despite having so many different types of on-device speech recognition models, each user only has at most one language pack per language, so such an API wouldn't add value for Chrome users. If we were to add the ability to select between different language packs, we would probably opt to make that a browser setting since that would be a decision made between the UA and the user rather than the website. 

At least for Chrome, I don't think we'll ever add the ability to select between language packs since the trend seems to be reducing the number of language packs. Back in 2020, we had different sized language packs for different device tiers, but as speech recognition models become more efficient and hardware specs improve, we're able to achieve low word error rates for a wider selection of users with fewer language packs. I wouldn't be surprised if in another 5 years we just have a single LLM capable of supporting all languages.

Even if other UAs want to allow users to select different language packs for different applications of the Web Speech API, it sounds like the hypothetical use case you described could still be accomplished without ever exposing the underlying model information/id to the website since the UA can prompt the user for anything before resolving the promises.

padenot: Yes, I would also favour option 2 for Firefox, and not expose all this complexity and fingerprinting surface to the Web, I was merely providing context about other specs.

The UA can pick a model that is suitable for the device and use-case on its own: offline transcription can use something slower/more accurate, `MediaStreamTrack` / microphone can use something that the UA knows will run at or faster than real-time speed on the device. The user can choose what models to install/uninstall as well (e.g. for us, it's in `about:inference`, but also in settings for other AI bits, etc.).

However, back to my initial point, we'll have Web developer asking for more options, and the fact that we're not providing those options needs to be clearly marked as intentional, probably in the `Privacy` section.

padenot: With that out of the way (barring any comment from e.g. @youenn), we can get back to API design.

First, let's not repeat the name of the API in method name. Second, we need consistency. We have:

> Promise<boolean> onDeviceWebSpeechAvailable(DOMString lang); 

and 

> Promise<boolean> installOnDeviceSpeechRecognition(DOMString lang);

why is the first saying Web Speech but the second only the recognition side of things? The former implies TTS as well.

It would look like this as things stand:

```js
var recognition = new SpeechRecognition();
let available = await recognition.onDeviceWebSpeechAvailable("fr-FR");
if (!available) {
  let installed = await recognition.installOnDeviceSpeechRecognition("fr-FR");
  if (!installed) {
    alert("User denied installation");
  }
}

assert(await recognition.onDecodeWebSpeechAvailable("fr-FR")); // this should always hold

// use the API
```

Why are those not static, with shorter, clearer names, without repetition?

```js
let available = await SpeechRecognition.availableOnDevice("fr-FR");
if (!available) {
  let installed = await SpeechRecognition.install("fr-FR");
  if (!installed) {
    alert("User denied installation");
  }
}

assert(await SpeechRecognition.isAvailable("fr-FR")); // this should always hold

// use the API
```

Still, it feels off to have this instead of something clearer like an `isSupported` method taking a language and a source.

evanbliu: I'm fine with those changes! What do you think about "installOnDevice"? The "OnDevice" part is kind of implied, but it mirrors "availableOnDevice" nicely and makes it explicit that it's for on-device speech recognition.

evanbliu: I've updated this PR with your suggestions. I don't have a strong preference for "installOnDevice" over "install", so if you strongly prefer "install" I'm fine with that too.

FYI, the proposed availability API for the [Web Translation API](https://github.com/webmachinelearning/translation-api) returns a enum with 4 possible values:  "unavailable", "downloadable", "downloading", "available". Does anyone have any opinions on whether the Web Speech API should also adopt this pattern?

evanbliu: @padenot - Do you have any other comments or suggestions on the current state of this PR?

index.bs
```diff
@@ -329,10 +329,10 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
-  <dd>The installOnDeviceSpeechRecognition method returns a boolean indicating whether the installation of on-device speech recognition for a given BCP 47 language tag initiated successfully. [[!BCP47]] Any website can automatically trigger a download of a new language pack if the user is connected to ethernet/Wifi and the language pack matches the user's preferred language. All sites must prompt the user if they wish to trigger a download over a cellular network or a language pack that does not match the user's preferred language.</dd>
+  <dd>The installOnDeviceSpeechRecognition method returns a Promise that resolves to a boolean indicating whether the installation of on-device speech recognition for a given BCP 47 language tag initiated successfully. [[!BCP47]] Any website can automatically trigger a download of a new language pack if the user is connected to ethernet/Wifi and the language pack matches the user's preferred language. All sites must prompt the user if they wish to trigger a download over a cellular network or a language pack that does not match the user's preferred language.</dd>
```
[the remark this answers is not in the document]
evanbliu: Sounds good! I'll remove the hand-wavy section for now and we can add constraints to the spec later, or alternatively just let UAs do their own thing.

PR #133, opened 2025-01-22
Add Explainer for On-Device Speech Recognition

evanbliu:
This PR adds an explainer for on-device speech recognition support.


ondevice/explainer.md
```diff
@@ -0,0 +1,76 @@
...
+SpeechRecognition.onDeviceWebSpeechAvailable(lang).then((available) => {
+    if (available) {
+        console.log(`On-device speech recognition is available for ${lang}.`);
+    } else {
+        console.log(`On-device speech recognition is not available for ${lang}.`);
+    }
+});
+```
+
+### 2. `installOnDeviceSpeechRecognition()`
```
chrisguttandin, line 41: I noticed that the return type is specified for the other method above. Perhaps it should be added here as well.

```suggestion
### 2. `Promise<undefined> installOnDeviceSpeechRecognition()`
```

ondevice/explainer.md
```diff
@@ -0,0 +1,76 @@
...
+## New Attribute
+
+### 1. `mode` attribute in the `SpeechRecognition` interface
+The `mode` attribute in the `SpeechRecognition` interface defines how speech recognition should behave when starting a session.
+
+#### `SpeechRecognitionMode` Enum
+
+- **"ondevice-preferred"**: Use on-device speech recognition if available. If not, fall back to cloud-based speech recognition.
+- **"ondevice-only"**: Only use on-device speech recognition. If it's unavailable, throw an error.
+- **"cloud-only"**: Only use cloud-based speech recognition, bypassing on-device options entirely.
```
domenic, line 62: I think these should be on-device-preferred and on-device-only. We don't have too many precedents on the web, but I can't see why we would remove the hyphens here.

ondevice/explainer.md
```diff
@@ -0,0 +1,76 @@
+# Explainer: On-Device Speech Recognition for the Web Speech API
+
+## Introduction
+
+The Web Speech API is a powerful browser feature that enables applications to perform speech recognition. Traditionally, this functionality relies on sending audio data to cloud-based services for recognition. While this approach is effective, it has certain drawbacks:
+
+- **Privacy concerns:** Raw and transcribed audio is transmitted over the network.
+- **Latency issues:** Users may experience delays due to network communication.
+- **Offline limitations:** Speech recognition does not work without an internet connection.
```
yashrajbharti, line 9:
```suggestion
- **Privacy concerns:** Raw and transcribed audio is transmitted over the network.
- **Latency issues:** Users may experience delays due to network communication.
- **Offline limitations:** Speech recognition does not work without an internet connection.
- **Sustainability concerns**: Excessive network requests and cloud dependency contribute to higher energy consumption and increased carbon emissions.
```

ondevice/explainer.md
```diff
@@ -0,0 +1,76 @@
...
+## Why Use On-Device Speech Recognition?
+
+### 1. **Privacy**
+On-device processing ensures that neither raw audio nor transcriptions leave the user's device, enhancing data security and user trust.
+
+### 2. **Performance**
+Local processing reduces latency, providing a smoother and faster user experience.
+
+### 3. **Offline Functionality**
+Applications can offer speech recognition capabilities even without an active internet connection, increasing their utility in remote or low-connectivity environments.
```
yashrajbharti, line 22:
```suggestion
### 1. **Privacy**
On-device processing ensures that neither raw audio nor transcriptions leave the user's device, enhancing data security and user trust.

### 2. **Performance**
Local processing reduces latency, providing a smoother and faster user experience.

### 3. **Offline Functionality**
Applications can offer speech recognition capabilities even without an active internet connection, increasing their utility in remote or low-connectivity environments.

### 4. **Sustainability and Energy Efficiency**
On-device speech recognition significantly reduces unnecessary network requests, which in turn helps minimize energy consumption. By using locally stored models, the solution reduces reliance on cloud-based services, leading to less data transfer and lower carbon emissions.
```

yashrajbharti:
Hi, this is an exciting approach! I've added a point on [sustainable web](https://websitesustainability.com/cache/files/research23.pdf) highlighting the environmental benefits of on-device processing.

I got to know from Kenji about the focus on supporting MediaTrack and on-device ASR. I wanted to confirm if this explainer also aims to cover those aspects, as I believe I could contribute here. If not, I wouldn't want to dilute the main message.

ondevice/explainer.md
yashrajbharti: Can further add:
This approach contributes to a more [sustainable web](https://websitesustainability.com/cache/files/research23.pdf), as it reduces both **latency** and **environmental impact**.

evanbliu:
Thanks for the suggestion @yashrajbharti! In this case, on-device speech recognition might not always have a reduced environmental impact. If a user rarely uses the Web Speech API, the bandwidth savings of on-device speech recognition might not outweigh the cost of downloading the on-device speech recognition models. Also, there it isn't exactly an apples-to-apples comparison between cloud and on-device speech recognition in terms of energy consumption and there are many factors that affect this. In Chrome for example, the cloud service that powers speech recognition is a bit antiquated but runs in highly optimized data centers whereas the local on-device speech recognition runs newer cutting edge models but has varying power consumption levels depending on the user's device. Because this isn't such a clear cut advantage, I suggest omitting it as an explicit benefit of using on-device speech recognition.

yashrajbharti:
Thanks for the detailed insight @evanbliu! While reducing network requests generally contributes to sustainability, I understand that the impact can vary. Without data to support average estimates, drawing a clear-cut advantage would indeed be difficult.

evanbliu:
@hoch - Can you please take a look at this PR when you get a chance? Thanks!

hoch (approved):
LGTM

Perhaps we can restructure the directory/file?

`explainers/on-device-speech-recognition.md`

Or are you planning to have more supporting materials for the on-device feature?

PR #143, opened 2025-02-20
Update the on-device speech recognition methods

### LINKED ISSUE #141: Consider changing return type of availableOnDevice()
In the current spec, availableOnDevice(lang) returns a Promise that resolves to a boolean indicating whether on-device speech recognition is supported.

The [Web Translation API](https://github.com/webmachinelearning/translation-api), which uses on-device language packs to perform translations, has an analogous API that uses a enum to indicate the language pack availability status rather than a boolean. The enum consists of the values: "unavailable", "downloadable", "downloading", and "available".

evanbliu:
Closes #141

This PR contains the following changes:
 - Introduce a new AvailabilityStatus enum to indicate the availability of on-device speech recognition for a given language.
 - Update the availableOnDevice method to return a Promise that resolves to an AvailabilityStatus instead of a boolean.
 - Rewrite the availableOnDevice and installOnDevice methods in the modern, algorithmic style.

***
<a href="https://pr-preview.s3.amazonaws.com/evanbliu/speech-api/pull/143.html" title="Last updated on Mar 31, 2025, 6:07 PM UTC (da5f09e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/143/6356249...evanbliu:da5f09e.html" title="Last updated on Mar 31, 2025, 6:07 PM UTC (da5f09e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -322,26 +322,26 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
 
   <dt><dfn method for=SpeechRecognition>availableOnDevice({{DOMString}} lang)</dfn> method</dt>
   <dd>
-    The availableOnDevice method returns a Promise that resolves to a {{AvailabilityStatus}} indicating the on-device speech recognition availability for a given [[!BCP47]] language tag.
+    The {{SpeechRecognition/availableOnDevice}} method returns a {{Promise}} that resolves to a {{AvailabilityStatus}} indicating the on-device speech recognition availability for a given [[!BCP47]] language tag.
 
     When invoked, run these steps:
     1. Let <var>promise</var> be <a>a new promise</a>.
-    1. Return <var>promise</var> and perform the remaining steps in parallel:
-    1. Run the <a>on-device availability algorithm</a> with <var>lang</var> and resolve <var>promise</var> with the result of the algorithm.
+    1. Run the <a>on-device availability algorithm</a> with <var>lang</var>. If it returns an exception, throw it and abort these steps.
+    1. Return <var>promise</var> and resolve it with the result of the <a>on-device availability algorithm</a>.
   </dd>
 
   <dt><dfn method for=SpeechRecognition>installOnDevice({{DOMString}} lang)</dfn> method</dt>
   <dd>
-    The installOnDevice method returns a Promise that resolves to a boolean indicating whether the installation of on-device speech recognition for a given [[!BCP47]] language tag initiated successfully.
+    The {{SpeechRecognition/installOnDevice}} method returns a {{Promise}} that resolves to a {{boolean}} indicating whether the installation of on-device speech recognition for a given [[!BCP47]] language tag initiated successfully.
 
     When invoked, run these steps:
     1. Let <var>promise</var> be <a>a new promise</a>.
-    1. Return <var>promise</var> and perform the remaining steps in parallel:
-    1. Run the <a>on-device availability algorithm</a> with <var>lang</var>.
+    1. Run the <a>on-device availability algorithm</a> with <var>lang</var>. If it returns an exception, throw it and abort these steps.
+    1. Return <var>promise</var>.
     1. If the <a>on-device availability algorithm</a> returns {{AvailabilityStatus/unavailable}}, {{AvailabilityStatus/downloading}}, or {{AvailabilityStatus/available}}, resolve <var>promise</var> with <code>false</code> and skip the rest of these steps.
     1. Initiate the download of the on-device speech recognition language for <var>lang</var>.
         <p class=note>
-          Note: The user agent may prompt the user for explicit permission to download the on-device speech recognition language pack.
+          Note: The user agent can prompt the user for explicit permission to download the on-device speech recognition language pack.
         </p>
     1. If the initiation of the download succeeds, resolve <var>promise</var> with <code>true</code>, otherwise resolve it with <code>false</code>.
   </dd>
@@ -350,10 +350,11 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
 <p>When the  <dfn>on-device availability algorithm</dfn> with <var>lang</var> is invoked, the user agent MUST run the following steps:
 1. If the [=current settings object=]'s [=relevant global object=]'s [=associated Document=] is NOT [=fully active=], throw an {{InvalidStateError}} and abort these steps.
 1. If <var>lang</var> is not a valid [[!BCP47]] language tag, throw a {{SyntaxError}} and abort these steps.
-1. If the on-device speech recognition language pack for <var>lang</var> is unsupported, return {{AvailabilityStatus/unavailable}}.
-1. If the on-device speech recognition language pack for <var>lang</var> is supported but not installed, return {{AvailabilityStatus/downloadable}}.
-1. If the on-device speech recognition language pack for <var>lang</var> is downloading, return {{AvailabilityStatus/downloading}}.
-1. If the on-device speech recognition language pack for <var>lang</var> is installed, return {{AvailabilityStatus/available}}.
+1. In parallel, do the following steps:
+    - If the on-device speech recognition language pack for <var>lang</var> is unsupported, return {{AvailabilityStatus/unavailable}}.
+    - If the on-device speech recognition language pack for <var>lang</var> is supported but not installed, return {{AvailabilityStatus/downloadable}}.
+    - If the on-device speech recognition language pack for <var>lang</var> is downloading, return {{AvailabilityStatus/downloading}}.
+    - If the on-device speech recognition language pack for <var>lang</var> is installed, return {{AvailabilityStatus/available}}.
 
 <p>When the <dfn>start session algorithm</dfn> with <var>requestMicrophonePermission</var> is invoked, the user agent MUST run the following steps:
 
--- a/index.bs
+++ b/index.bs
@@ -261,7 +261,7 @@ dictionary SpeechRecognitionEventInit : EventInit {
 
   <dt><dfn attribute for=SpeechRecognition>lang</dfn> attribute</dt>
   <dd>This attribute will set the language of the recognition for the request, using a valid BCP 47 language tag. [[!BCP47]]
-  If unset it remains unset for getting in script, but will default to use the <a spec=html>language</a> of the html document root element and associated hierarchy.
+  If unset it remains unset for getting in script, but will default to use the language of the html document root element and associated hierarchy.
   This default value is computed and used when the input request opens a connection to the recognition service.</dd>
 
   <dt><dfn attribute for=SpeechRecognition>continuous</dfn> attribute</dt>
@@ -704,7 +704,7 @@ interface SpeechSynthesisVoice {
 
   <dt><dfn attribute for=SpeechSynthesisUtterance>lang</dfn> attribute</dt>
   <dd>This attribute specifies the language of the speech synthesis for the utterance, using a valid BCP 47 language tag. [[!BCP47]]
-  If unset it remains unset for getting in script, but will default to use the <a spec=html>language</a> of the html document root element and associated hierarchy.
+  If unset it remains unset for getting in script, but will default to use the language of the html document root element and associated hierarchy.
   This default value is computed and used when the input request opens a connection to the recognition service.</dd>
 
   <dt><dfn attribute for=SpeechSynthesisUtterance>voice</dfn> attribute</dt>
--- a/index.bs
+++ b/index.bs
@@ -326,8 +326,8 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
 
     When invoked, run these steps:
     1. Let <var>promise</var> be <a>a new promise</a>.
-    1. Run the <a>on-device availability algorithm</a> with <var>lang</var>. If it returns an exception, throw it and abort these steps.
-    1. Return <var>promise</var> and resolve it with the result of the <a>on-device availability algorithm</a>.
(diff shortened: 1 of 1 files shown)


evanbliu:
@padenot - Can you please take a look at this PR when you get a chance? Thanks!

index.bs
```diff
@@ -314,12 +321,39 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+    1. Initiate the download of the on-device speech recognition language for <var>lang</var>.
+        <p class=note>
+          Note: The user agent may prompt the user for explicit permission to download the on-device speech recognition language pack.
+        </p>
+    1. If the initiation of the download succeeds, resolve <var>promise</var> with <code>true</code>, otherwise resolve it with <code>false</code>.
+  </dd>
 
 </dl>
+<p>When the  <dfn>on-device availability algorithm</dfn> with <var>lang</var> is invoked, the user agent MUST run the following steps:
+1. If the [=current settings object=]'s [=relevant global object=]'s [=associated Document=] is NOT [=fully active=], throw an {{InvalidStateError}} and abort these steps.
```
padenot, line 351: You can't do this because you're in parallel. You need to do this check synchronously, thrown at this moment, then do the rest in parallel.

We can do it like so, in the algorithm:

1. Check if doc is fully active, if not, return a DOM Exception of type ... and abort these steps
2. Check if lang is valid per bcp47. if not, return a DOM Exception of type ... and abort these steps
3. In parallel, do the following steps:
    1. step a
    2. step b
    3. step c
    4. step d
    5. Queue a task back to the main thread to resolve the promise with the value we've determined in step a/b/c/d

And above at the call site of the algorithm:

1. Let p be a new promise
2. Run the algorithm. If it returns an exception, throw it and abort these steps
3. Return p

index.bs
```diff
@@ -314,12 +321,39 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+    When invoked, run these steps:
+    1. Let <var>promise</var> be <a>a new promise</a>.
+    1. Return <var>promise</var> and perform the remaining steps in parallel:
+    1. Run the <a>on-device availability algorithm</a> with <var>lang</var> and resolve <var>promise</var> with the result of the algorithm.
+  </dd>
 
   <dt><dfn method for=SpeechRecognition>installOnDevice({{DOMString}} lang)</dfn> method</dt>
-  <dd>The installOnDevice method returns a Promise that resolves to a boolean indicating whether the installation of on-device speech recognition for a given BCP 47 language tag initiated successfully. [[!BCP47]]</dd>
+  <dd>
+    The installOnDevice method returns a Promise that resolves to a boolean indicating whether the installation of on-device speech recognition for a given [[!BCP47]] language tag initiated successfully.
```
padenot, line 335: Similarly, use markup to autolink.

index.bs
```diff
@@ -314,12 +321,39 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
   If the abort method is called on an object which is already stopped or aborting (that is, start was never called on it, the <a event for=SpeechRecognition>end</a> or <a event for=SpeechRecognition>error</a> event has fired on it, or abort was previously called on it), the user agent must ignore the call.</dd>
 
   <dt><dfn method for=SpeechRecognition>availableOnDevice({{DOMString}} lang)</dfn> method</dt>
-  <dd>The availableOnDevice method returns a Promise that resolves to a boolean indicating whether on-device speech recognition is available for a given BCP 47 language tag. [[!BCP47]]</dd>
+  <dd>
+    The availableOnDevice method returns a Promise that resolves to a {{AvailabilityStatus}} indicating the on-device speech recognition availability for a given [[!BCP47]] language tag.
```
padenot, line 325: Markup around `availableOnDevice`. After merging my PR, it could simply be:

```
{{SpeechRecognition/installOnDevice}}
```

this allows linking back to the IDL.

Similarly for promise, boolean, etc.

index.bs
```diff
@@ -314,12 +321,39 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+    The installOnDevice method returns a Promise that resolves to a boolean indicating whether the installation of on-device speech recognition for a given [[!BCP47]] language tag initiated successfully.
+
+    When invoked, run these steps:
+    1. Let <var>promise</var> be <a>a new promise</a>.
+    1. Return <var>promise</var> and perform the remaining steps in parallel:
+    1. Run the <a>on-device availability algorithm</a> with <var>lang</var>.
+    1. If the <a>on-device availability algorithm</a> returns {{AvailabilityStatus/unavailable}}, {{AvailabilityStatus/downloading}}, or {{AvailabilityStatus/available}}, resolve <var>promise</var> with <code>false</code> and skip the rest of these steps.
+    1. Initiate the download of the on-device speech recognition language for <var>lang</var>.
+        <p class=note>
+          Note: The user agent may prompt the user for explicit permission to download the on-device speech recognition language pack.
```
padenot, line 344: You cannot use `MAY`, which is an RFC 2119 keyword that has normative conformance class, in a `Note`, which is by definition informative (in other words, non-normative).

I'll turn the bikeshed option to warn about this shortly, I should have done it already, sorry.

We can simply rephrase here, say "can", that often works and is similar in meaning for what we want to say.

index.bs
```diff
@@ -314,12 +321,39 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+  </dd>
 
 </dl>
+<p>When the  <dfn>on-device availability algorithm</dfn> with <var>lang</var> is invoked, the user agent MUST run the following steps:
+1. If the [=current settings object=]'s [=relevant global object=]'s [=associated Document=] is NOT [=fully active=], throw an {{InvalidStateError}} and abort these steps.
+1. If <var>lang</var> is not a valid [[!BCP47]] language tag, throw a {{SyntaxError}} and abort these steps.
+1. If the on-device speech recognition language pack for <var>lang</var> is unsupported, return {{AvailabilityStatus/unavailable}}.
+1. If the on-device speech recognition language pack for <var>lang</var> is supported but not installed, return {{AvailabilityStatus/downloadable}}.
+1. If the on-device speech recognition language pack for <var>lang</var> is downloading, return {{AvailabilityStatus/downloading}}.
+1. If the on-device speech recognition language pack for <var>lang</var> is installed, return {{AvailabilityStatus/available}}.
```
padenot, line 356: When do we know it has finished downloading?

index.bs
```diff
@@ -200,6 +200,13 @@ enum SpeechRecognitionMode {
     "cloud-only", // Cloud speech recognition only.
 };
 
+enum AvailabilityStatus {
+    "unavailable",
+    "downloadable",
+    "downloading",
+    "available"
+};
```
padenot, line 208: I was thinking maybe we could reuse Translations' enum but maybe there's no point.

padenot (changes requested):

index.bs
```diff
@@ -314,12 +321,40 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
 
 </dl>
+<p>When the  <dfn>on-device availability algorithm</dfn> with <var>lang</var> is invoked, the user agent MUST run the following steps:
+1. If the [=current settings object=]'s [=relevant global object=]'s [=associated Document=] is NOT [=fully active=], throw an {{InvalidStateError}} and abort these steps.
+1. If <var>lang</var> is not a valid [[!BCP47]] language tag, throw a {{SyntaxError}} and abort these steps.
+1. In parallel, do the following steps:
+    - If the on-device speech recognition language pack for <var>lang</var> is unsupported, return {{AvailabilityStatus/unavailable}}.
+    - If the on-device speech recognition language pack for <var>lang</var> is supported but not installed, return {{AvailabilityStatus/downloadable}}.
+    - If the on-device speech recognition language pack for <var>lang</var> is downloading, return {{AvailabilityStatus/downloading}}.
+    - If the on-device speech recognition language pack for <var>lang</var> is installed, return {{AvailabilityStatus/available}}.
```
padenot, line 357: We need to add steps to queue a task back to the main thread here, with the value decided by the four lines. In this task, we can resolve the promise with this value.

Because of this, this algorithm needs to take an additional parameter, the promise. The queue in which the task is enqueued is inferred by the fact that we're checking on an `associated Document` etc., so generally we're not too specific.

padenot (changes requested):
Way better, still some comments.

How does one know that the model has finished downloaded?

evanbliu:
> Way better, still some comments.
> 
> How does one know that the model has finished downloaded?

From above:
Right now sites will have to poll availableOnDevice to find out if the download has finished.

What are your thoughts on either:

Introducing a new languageinstalled SpeechRecognition event.
Introducing a new API that returns a Promise that resolves to true once the language is finished or resolves to false if the installation failed.
Updating installOnDevice to resolve to true once on-device speech recognition is installed rather than resolving to true when the installation is initiated.
Just require websites to poll availableOnDevice.

beaufortfrancois:
FYI https://github.com/webmachinelearning/prompt-api?tab=readme-ov-file#download-progress may be of interest for this group.

evanbliu:
Gentle ping -- @padenot - Do you have any additional comments on this CL, or on how the API for detecting if the download has finished should look like?

evanbliu:
> Gentle ping -- @padenot - Do you have any additional comments on this CL, or on how the API for detecting if the download has finished should look like?

I'd like to propose option 3: update installOnDevice to resolve to true once on-device speech recognition is installed rather than resolving to true when the installation is initiated.

The Promise will resolve to false if the download fails for any reason (i.e. user rejects the download, network issues, unsupported language) and resolve to true once the language is installed and ready. I don't think the delay should be much of an issue because in most cases the download would complete relatively quickly. Of course, there will be some cases in which the download takes much longer, but the current installOnDevice API could also take a long time to resolve to True if the user takes awhile to respond to a prompt, so websites will need to handle this delay already anyways.

padenot:
Yes, I'd tend to agree. We could imagine having a UA-side progress indicator here.

evanbliu:
Sounds good, I've updated the spec to reflect this. Let me know if you have any other comments or concerns on this PR!

index.bs
```diff
@@ -314,12 +321,47 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+    1. If the on-device speech recognition language pack for <var>lang</var> is unsupported, return a resolved {{Promise}} with false and skip the rest of these steps.
+    1. Let <var>promise</var> be <a>a new promise</a>.
+    1. Initiate the download of the on-device speech recognition language for <var>lang</var>.
+        <p class=note>
+          Note: The user agent can prompt the user for explicit permission to download the on-device speech recognition language pack.
+        </p>
+    1. [=Queue a task=] on the [=relevant global object=]'s [=task queue=] to run the following step:
+        - If the download succeeds, resolve <var>promise</var> with <code>true</code>, otherwise resolve it with <code>false</code>.
+            <p class="note">
+              Note: The <code>false</code> resolution of the Promise does not indicate the specific cause of failure. User agents are encouraged to provide more detailed information about the failure in developer tools console messages. However, this detailed error information is not exposed to the script. This is non-normative.
```
padenot, line 349: Notes are always non-normative by definition, you can drop the last sentence here.

index.bs
```diff
@@ -314,12 +321,47 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+            </p>
+    1. Return <var>promise</var>.
+  </dd>
 
 </dl>
+<p>When the  <dfn>on-device availability algorithm</dfn> with <var>lang</var> and <var>promise</var> is invoked, the user agent MUST run the following steps:
+1. If the [=current settings object=]'s [=relevant global object=]'s [=associated Document=] is NOT [=fully active=], throw an {{InvalidStateError}} and abort these steps.
+1. If <var>lang</var> is not a valid [[!BCP47]] language tag, throw a {{SyntaxError}} and abort these steps.
+1. In parallel, determine the availability status for <var>lang</var>:
```
padenot, line 358: Is there a particular reason we want to do this in parallel? When implementing, we likely have the value of this state synchronously available on the thread on which we need it, because any change to this state is caused by `installOnDevice`, and this changes the state on the thread itself.

Are we worried about UA specific bits of UI, or initial fetch of this value?

index.bs
```diff
@@ -314,12 +321,47 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+    When invoked, run these steps:
+    1. Let <var>promise</var> be <a>a new promise</a>.
+    1. Run the <a>on-device availability algorithm</a> with <var>lang</var> and <var>promise</var>. If it returns an exception, throw it and abort these steps.
+    1. Return <var>promise</var>.
+  </dd>
 
   <dt><dfn method for=SpeechRecognition>installOnDevice({{DOMString}} lang)</dfn> method</dt>
-  <dd>The installOnDevice method returns a Promise that resolves to a boolean indicating whether the installation of on-device speech recognition for a given BCP 47 language tag initiated successfully. [[!BCP47]]</dd>
+  <dd>
+    The {{SpeechRecognition/installOnDevice}} method returns a {{Promise}} that resolves to a {{boolean}} indicating whether the installation of on-device speech recognition for a given [[!BCP47]] language tag succeeded.
```
padenot, line 335: that resolves to a {{boolean}} `when and whether`

padenot (approved):
Looks good, some comments, but nothing too important. If the answer to the last sentence of the 3rd comment is yes, this is good for merging w/ the small adjustment I mentioned.

PR #145, opened 2025-02-24
Add speech recognition phrase list to the Web Speech API

yrw-google:
Introduce a new speech recognition phrases feature for contextual biasing

***
<a href="https://pr-preview.s3.amazonaws.com/yrw-google/web-speech-api/pull/145.html" title="Last updated on Apr 10, 2025, 8:50 PM UTC (2e4b76b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/145/bad1722...yrw-google:2e4b76b.html" title="Last updated on Apr 10, 2025, 8:50 PM UTC (2e4b76b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -151,6 +151,18 @@ The term "interim result" indicates a SpeechRecognitionResult in which the final
         A boolean flag representing whether the speech recognition started. The initial value is <code>false</code>.
 </dl>
 
+<dl dfn-type=attribute dfn-for="SpeechRecognition">
+    : <dfn>[[mode]]</dfn>
+    ::
+        A {{SpeechRecognitionMode}} enum to determine where speech recognition takes place. The initial value is <code>ondevice-preferred</code>.
+</dl>
+
+<dl dfn-type=attribute dfn-for="SpeechRecognition">
+    : <dfn>[[phrases]]</dfn>
+    ::
+        A {{SpeechRecognitionPhraseList}} representing a list of phrases for contextual biasing. The initial value is null.
+</dl>
+
 <xmp class="idl">
 [Exposed=Window]
 interface SpeechRecognition : EventTarget {
@@ -162,7 +174,7 @@ interface SpeechRecognition : EventTarget {
     attribute boolean interimResults;
     attribute unsigned long maxAlternatives;
     attribute SpeechRecognitionMode mode;
-    attribute SpeechRecognitionContext context;
+    attribute SpeechRecognitionPhraseList phrases;
 
     // methods to drive the speech interaction
     undefined start();
@@ -194,7 +206,7 @@ enum SpeechRecognitionErrorCode {
     "not-allowed",
     "service-not-allowed",
     "language-not-supported",
-    "context-not-supported"
+    "phrases-not-supported"
 };
 
 enum SpeechRecognitionMode {
@@ -265,27 +277,20 @@ interface SpeechRecognitionPhrase {
     readonly attribute float boost;
 };
 
-// The object representing a list of biasing phrases.
+// The object representing a list of phrases for contextual biasing.
 [Exposed=Window]
 interface SpeechRecognitionPhraseList {
-    constructor();
+    constructor(sequence<SpeechRecognitionPhrase> phrases);
     readonly attribute unsigned long length;
     SpeechRecognitionPhrase item(unsigned long index);
     undefined addItem(SpeechRecognitionPhrase item);
-};
-
-// The object representing a recognition context collection.
-[Exposed=Window]
-interface SpeechRecognitionContext {
-    constructor(SpeechRecognitionPhraseList phrases);
-    readonly attribute SpeechRecognitionPhraseList phrases;
+    undefined removeItem(unsigned long index);
 };
 </xmp>
 
 <h4 id="speechreco-attributes">SpeechRecognition Attributes</h4>
 
 <dl>
-
   <dt><dfn attribute for=SpeechRecognition>lang</dfn> attribute</dt>
   <dd>This attribute will set the language of the recognition for the request, using a valid BCP 47 language tag. [[!BCP47]]
   If unset it remains unset for getting in script, but will default to use the language of the html document root element and associated hierarchy.
@@ -309,10 +314,34 @@ interface SpeechRecognitionContext {
   The default value is 1.</dd>
 
   <dt><dfn attribute for=SpeechRecognition>mode</dfn> attribute</dt>
-  <dd>An enum to determine where speech recognition takes place. The default value is "ondevice-preferred".</dd>
+  <dd>
+   This attribute represents where speech recognition takes place.
+  </dd>
+  <dd>
+    The getter steps are to return the value of {{SpeechRecognition/[[mode]]}}.
+  </dd>
+  <dd>
+    The setter steps are:
+    1. If the {{SpeechRecognitionPhraseList/length}} of {{SpeechRecognition/phrases}} is greater than 0
+        and the system using the given value for {{SpeechRecognition/[[mode]]}} does not support contextual biasing,
+        throw a {{SpeechRecognitionErrorEvent}} with the {{SpeechRecognitionErrorCode/not-allowed}} error code and abort these steps.
+    1. Set {{SpeechRecognition/[[mode]]}} to the given value.
+  </dd>
 
-  <dt><dfn attribute for=SpeechRecognition>context</dfn> attribute</dt>
-  <dd>This attribute will set the speech recognition context for the recognition session to start with.</dd>
+  <dt><dfn attribute for=SpeechRecognition>phrases</dfn> attribute</dt>
+  <dd>
+    This attribute represents a list of phrases for contextual biasing.
+  </dd>
+  <dd>
+    The getter steps are to return the value of {{SpeechRecognition/[[phrases]]}}.
+  </dd>
+  <dd>
+    The setter steps are:
+    1. If the {{SpeechRecognitionPhraseList/length}} of the given value is greater than 0 and the system does not support contextual biasing,
+        throw a {{SpeechRecognitionErrorEvent}} with the {{phrases-not-supported}} error code and abort these steps.
+    1. Set {{SpeechRecognition/[[phrases]]}} to the given value.
+    1. Send a copy of {{SpeechRecognition/[[phrases]]}} to the system for initializing or updating the phrases for contextual biasing implementation.
+  </dd>
 </dl>
 
 <p class=issue>The group has discussed whether WebRTC might be used to specify selection of audio sources and remote recognizers.
@@ -357,17 +386,6 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
   When the abort method is called, the speech service must stop recognizing.
   The user agent must raise an <a event for=SpeechRecognition>end</a> event once the speech service is no longer connected.
   If the abort method is called on an object which is already stopped or aborting (that is, start was never called on it, the <a event for=SpeechRecognition>end</a> or <a event for=SpeechRecognition>error</a> event has fired on it, or abort was previously called on it), the user agent must ignore the call.</dd>
-
-  <dt><dfn method for=SpeechRecognition>updateContext({{SpeechRecognitionContext}} |context|)</dfn> method</dt>
-  <dd>
-    The updateContext method updates the speech recognition context after the speech recognition session has started.
-    If the session has not started yet, user should update {{SpeechRecognition/context}} instead of using this method.
-
-    When invoked, run the following steps:
-    1. If {{[[started]]}} is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
-    1. If the system does not support speech recognition context, throw a {{SpeechRecognitionErrorEvent}} with the {{context-not-supported}} error code and abort these steps.
-    1. The system updates its speech recognition context to be |context|.
-  </dd>
 
   <dt><dfn method for=SpeechRecognition>availableOnDevice({{DOMString}} lang)</dfn> method</dt>
   <dd>
@@ -426,9 +444,6 @@ following steps:
 1. If |requestMicrophonePermission| is `true` and [=request
     permission to use=] "`microphone`" is [=permission/"denied"=], abort
     these steps.
-1. If {{SpeechRecognition/context}} is not null and the system does not support
-    speech recognition context, throw a {{SpeechRecognitionErrorEvent}} with the
-    {{context-not-supported}} error code and abort these steps.
 1. Once the system is successfully listening to the recognition, queue a task to
     [=fire an event=] named <a event for=SpeechRecognition>start</a> at [=this=].
 
@@ -523,8 +538,8 @@ For example, some implementations may fire <a event for=SpeechRecognition>audioe
     <dt><dfn enum-value for=SpeechRecognitionErrorCode>"language-not-supported"</dfn></dt>
     <dd>The language was not supported.</dd>
 
-    <dt><dfn enum-value for=SpeechRecognitionErrorCode>"context-not-supported"</dfn></dt>
-    <dd>The speech recognition model does not support speech recognition context.</dd>
+    <dt><dfn enum-value for=SpeechRecognitionErrorCode>"phrases-not-supported"</dfn></dt>
+    <dd>The speech recognition model does not support phrases for contextual biasing.</dd>
   </dl>
   </dd>
 
@@ -605,62 +620,91 @@ For a non-continuous recognition it will hold only a single value.</p>
 
 <h4 id="speechreco-phrase">SpeechRecognitionPhrase</h4>
 
-<p>The SpeechRecognitionPhrase object represents a phrase for contextual biasing.</p>
+<p>The SpeechRecognitionPhrase object represents a phrase for contextual biasing and has the following internal slots:</p>
+
+<dl dfn-type=attribute dfn-for="SpeechRecognitionPhrase">
+    : <dfn>[[phrase]]</dfn>
+    ::
+        A {{DOMString}} representing the text string to be boosted. The initial value is null.
+</dl>
+
+<dl dfn-type=attribute dfn-for="SpeechRecognitionPhrase">
+    : <dfn>[[boost]]</dfn>
+    ::
+        A float representing approximately the natural log of the number of times more likely the website thinks this phrase is
+        than what the speech recognition model knows.
+        A valid boost must be a float value inside the range [0.0, 10.0], with a default value of 1.0 if not specified.
+        A boost of 0.0 means the phrase is not boosted at all, and a higher boost means the phrase is more likely to appear.
+        A boost of 10.0 means the phrase is extremely likely to appear and should be rarely set.
+</dl>
 
 <dl>
   <dt><dfn constructor for=SpeechRecognitionPhrase>SpeechRecognitionPhrase(|phrase|, |boost|)</dfn> constructor</dt>
   <dd>
-    When invoked, run the following steps:
-    1. If the |phrase| is an empty string, throw a "{{SyntaxError}}" {{DOMException}}.
-    1. If the |boost| is smaller than 0.0 or greater than 10.0, throw a "{{SyntaxError}}" {{DOMException}}.
-    1. Construct a new SpeechRecognitionPhrase object with |phrase| and |boost|.
-    1. Return the object.
+    When this constructor is invoked, run the following steps:
+    1. If |phrase| is an empty string, throw a {{SyntaxError}} and abort these steps.
+    1. If |boost| is smaller than 0.0 or greater than 10.0, throw a {{SyntaxError}} and abort these steps.
+    1. Let |phr| be a new object of type {{SpeechRecognitionPhrase}}.
+    1. Set |phr|.{{[[phrase]]}} to be the value of |phrase|.
+    1. Set |phr|.{{[[boost]]}} to be the value of |boost|.
+    1. Return |phr|.
   </dd>
 
   <dt><dfn attribute for=SpeechRecognitionPhrase>phrase</dfn> attribute</dt>
-  <dd>This attribute is the text string to be boosted.</dd>
+  <dd>This attribute returns the value of {{[[phrase]]}}.</dd>
 
   <dt><dfn attribute for=SpeechRecognitionPhrase>boost</dfn> attribute</dt>
-  <dd>This attribute is approximately the natural log of the number of times more likely the website thinks this phrase is than what the speech recognition model knows.
-  A valid boost must be a float value inside the range [0.0, 10.0], with a default value of 1.0 if not specified.
-  A boost of 0.0 means the phrase is not boosted at all, and a higher boost means the phrase is more likely to appear.
-  A boost of 10.0 means the phrase is extremely likely to appear and should be rarely set.
-  </dd>
+  <dd>This attribute returns the value of {{[[boost]]}}.</dd>
 </dl>
 
 <h4 id="speechreco-phraselist">SpeechRecognitionPhraseList</h4>
 
-<p>The SpeechRecognitionPhraseList object holds a sequence of phrases for contextual biasing.</p>
+<p>The SpeechRecognitionPhraseList object holds a sequence of phrases for contextual biasing and has the following internal slot:</p>
+
+<dl dfn-type=attribute dfn-for="SpeechRecognitionPhraseList">
+    : <dfn>[[phrases]]</dfn>
+    ::
+        A sequence of {{SpeechRecognitionPhrase}} representing the phrases to be boosted. The initial value is an empty list.
+</dl>
 
 <dl>
-  <dt><dfn constructor for=SpeechRecognitionPhraseList>SpeechRecognitionPhraseList()</dfn> constructor</dt>
-  <dd>This constructor returns an empty list.</dd>
+  <dt><dfn constructor for=SpeechRecognitionPhraseList>SpeechRecognitionPhraseList(|phrases|)</dfn> constructor</dt>
+  <dd>
+    When this constructor is invoked, run the following steps:
+    1. Let |list| be a new object of type {{SpeechRecognitionPhraseList}}.
+    1. Set |list|.{{SpeechRecognitionPhraseList/[[phrases]]}} to be the value of |phrases|.
+    1. Return |list|.
+  </dd>
 
   <dt><dfn attribute for=SpeechRecognitionPhraseList>length</dfn> attribute</dt>
-  <dd>This attribute indicates how many phrases are in the list. The user agent must ensure it is set to the number of phrases in the list.</dd>
-
-  <dt><dfn method for=SpeechRecognitionPhraseList>SpeechRecognitionPhrase(|index|)</dfn> method</dt>
   <dd>
-    This method gets the SpeechRecognitionPhrase object at the |index| of the list.
+    This attribute indicates the number of phrases in the list.
+    When invoked, return the number of items in {{SpeechRecognitionPhraseList/[[phrases]]}}.
+  </dd>
+
+  <dt><dfn method for=SpeechRecognitionPhraseList>item(|index|)</dfn> method</dt>
+  <dd>
+    This method gets the {{SpeechRecognitionPhrase}} object at the |index| of the list.
     When invoked, run the following steps:
-    1. If the |index| is smaller than 0, or greater than or equal to {{SpeechRecognitionPhraseList/length}}, return null.
-    1. Return the SpeechRecognitionPhrase from the |index| of the list.
+    1. If |index| is smaller than 0, or greater than or equal to {{SpeechRecognitionPhraseList/length}},
+        throw a {{RangeError}} and abort these steps.
+    1. Return the {{SpeechRecognitionPhrase}} at the |index| of {{SpeechRecognitionPhraseList/[[phrases]]}}.
   </dd>
 
   <dt><dfn method for=SpeechRecognitionPhraseList>addItem(|item|)</dfn> method</dt>
-  <dd>This method adds the SpeechRecognitionPhrase object |item| to the end of the list.</dd>
-</dl>
+  <dd>
+    This method adds the {{SpeechRecognitionPhrase}} object |item| to the list.
+    When invoked, add |item| to the end of {{SpeechRecognitionPhraseList/[[phrases]]}}.
+  </dd>
 
-<h4 id="speechreco-context">SpeechRecognitionContext</h4>
-
-<p>The SpeechRecognitionContext object holds contextual information to provide to the speech recognition models.</p>
-
-<dl>
-  <dt><dfn constructor for=SpeechRecognitionContext>SpeechRecognitionContext(|phrases|)</dfn> constructor</dt>
-  <dd>This constructor returns a new SpeechRecognitionContext object with the SpeechRecognitionPhraseList object |phrases| in it.</dd>
-
-  <dt><dfn attribute for=SpeechRecognitionContext>phrases</dfn> attribute</dt>
-  <dd>This attribute represents the phrases to be boosted.</dd>
+  <dt><dfn method for=SpeechRecognitionPhraseList>removeItem(|index|)</dfn> method</dt>
+  <dd>
+    This method removes the {{SpeechRecognitionPhrase}} object at the |index| of the list.
+    When invoked, run the following steps:
+    1. If |index| is smaller than 0, or greater than or equal to {{SpeechRecognitionPhraseList/length}},
+        throw a {{RangeError}} and abort these steps.
+    1. Remove the {{SpeechRecognitionPhrase}} object at the |index| of {{SpeechRecognitionPhraseList/[[phrases]]}}.
+  </dd>
 </dl>
 
 <h3 id="tts-section">The SpeechSynthesis Interface</h3>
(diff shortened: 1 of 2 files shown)


yrw-google:
Hi @padenot, can you review this one too and also take a look at the explainer again? Feel free to assign another reviewer too if needed

index.bs
```diff
@@ -246,6 +249,28 @@ dictionary SpeechRecognitionEventInit : EventInit {
     unsigned long resultIndex = 0;
     required SpeechRecognitionResultList results;
 };
+
+// The object representing a biasing phrase.
+[Exposed=Window]
+interface SpeechRecognitionPhrase {
+    // If the phrase is empty or the boost is outside the range [0, 10], throw a "SyntaxError" DOMException.
```
padenot, line 256: Generally, no comments in IDL. Instead, normatively specify below with an algorithm.

Why is it [0, 10] and not [0, 1].

index.bs
```diff
@@ -246,6 +249,28 @@ dictionary SpeechRecognitionEventInit : EventInit {
...
+    constructor(DOMString phrase, optional float boost = 1.0);
```
padenot, line 257: Optional is not needed of there's a default value iirc (please double check).

index.bs
```diff
@@ -246,6 +249,28 @@ dictionary SpeechRecognitionEventInit : EventInit {
...
+    attribute DOMString phrase;
+    attribute float boost;
```
padenot, line 259: Is this `readonly`, since it's set in the ctor?

Is the boost really a float, or an integer in [0, 10]?

index.bs
```diff
@@ -246,6 +249,28 @@ dictionary SpeechRecognitionEventInit : EventInit {
...
+};
+
+// The object representing a list of biasing phrases.
+[Exposed=Window]
+interface SpeechRecognitionPhraseList {
+    readonly attribute unsigned long length;
+    getter SpeechRecognitionPhrase item(unsigned long index);
```
padenot, line 266: This is invalid WebIDL.

It would be either:

```webidl
SpeechRecognition item(unsigned long index);
```

or

```webidl
getter SpeechRecognition(unsigned long index)
```

What is the intent here?

index.bs
```diff
@@ -246,6 +249,28 @@ dictionary SpeechRecognitionEventInit : EventInit {
...
+};
+
+// The object representing a recognition context collection.
+[Exposed=Window]
+interface SpeechRecognitionContext {
+    attribute SpeechRecognitionPhraseList phrases;
```
padenot, line 272: Is this `readonly` ? Can we replace the `phrase` attribute by another one?

index.bs
```diff
@@ -276,6 +301,10 @@ dictionary SpeechRecognitionEventInit : EventInit {
 
   <dt><dfn attribute for=SpeechRecognition>mode</dfn> attribute</dt>
   <dd>An enum to determine where speech recognition takes place. The default value is "ondevice-preferred".</dd>
+
+  <dt><dfn attribute for=SpeechRecognition>context</dfn> attribute</dt>
+  <dd>This attribute will set the speech recognition context for the recognition session to start with.
+  If the speech recognition model does not support recognition context, a {{SpeechRecognitionErrorEvent}} with the {{recognition-context-not-supported}} error code will be thrown.</dd>
```
padenot, line 307: Please use algorithmic form for new members, see e.g. https://github.com/WebAudio/web-speech-api/pull/144, we're converting the spec to that.

index.bs
```diff
@@ -313,12 +342,15 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
   The user agent must raise an <a event for=SpeechRecognition>end</a> event once the speech service is no longer connected.
   If the abort method is called on an object which is already stopped or aborting (that is, start was never called on it, the <a event for=SpeechRecognition>end</a> or <a event for=SpeechRecognition>error</a> event has fired on it, or abort was previously called on it), the user agent must ignore the call.</dd>
 
+  <dt><dfn method for=SpeechRecognition>updateContext({{SpeechRecognitionContext}} context)</dfn> method</dt>
+  <dd>The updateContext method updates the speech recognition context after the speech recognition session has started. If the recognition session is not active when this method is called, throw an {{InvalidStateError}} and abort.
+  If the speech recognition model does not support recognition context, throw a {{SpeechRecognitionErrorEvent}} with the {{recognition-context-not-supported}} error code and abort.</dd>
```
padenot, line 347: Similarly, rewrite in modern style, so it is precisely defined when this works, instead of in prose.

Why prevent setting it beforehand? We can simply make the implementation buffer it and apply it on start?

padenot (changes requested):

index.bs
```diff
@@ -191,7 +193,8 @@ enum SpeechRecognitionErrorCode {
     "network",
     "not-allowed",
     "service-not-allowed",
-    "language-not-supported"
+    "language-not-supported",
+    "recognition-context-not-supported"
```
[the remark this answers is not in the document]
yrw-google: I think we can remove "recognition" since the other error codes don't have it, and as long as we don't use the word "context" for other usages it should be fine.

index.bs
yrw-google: I've removed this comment and added a paragraph for SpeechRecognitionPhrase with an algorithm for the constructor. I also explained why it's [0, 10] for the boost attribute.

index.bs
yrw-google: If I remove "optional", bikeshed will throw an error so I think it's required.

index.bs
yrw-google: I've marked the attributes as read only for now, so they can only be set in the constructor. The boost can be a float number in our speech recognition model so we allow it.

index.bs
yrw-google: I've changed this to `SpeechRecognitionPhrase item(unsigned long index)`, but using getter is actually how `SpeechRecognitionResultList` is doing it, as well as some list objects I've seen in other specs, e.g. https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#the-domstringlist-interface. I thought it's a standard thing to always define a getter for a list, but I don't see that getter is required in our use case, so I can either keep it or remove it.

index.bs
yrw-google: I've marked this as `readonly`.

index.bs
yrw-google: I've removed the exception here and added it in the start session algorithm (step 5).

index.bs
yrw-google: I've added an algorithm for this. I think we should not allow this to be set beforehand, since that will conflict with what the user has set in {{SpeechRecognition/context}}. If user calls `updateContext()` and then sets {{SpeechRecognition/context}} again, it will be a mess. I think the goal of `updateContext()`is to provide a way to update context after the session has started. If we don't have it, user needs to set {{SpeechRecognition/context}} and call `start()` again to update context, which is not ideal.

yrw-google:
Hi @padenot, I've updated the specs as well as the explainer according to your comments. Please take a look again when you get a chance. Thanks!

padenot:
Can we please focus on either the explainer or the spec patch? If we have a spec patch, an explainer shouldn't be necessary. If we aren't comfortable writing the spec patch right now because we want to iterate, it doesn't seem useful to update the spec patch.

Let me know which one I should look at first please?

yrw-google:
Hi @padenot, you can focus on the spec patch right now. I'm keeping the spec patch and the explainer in sync and the spec patch has many more details than the explainer, so if we can reach consensus on the spec, we can also reach consensus on the explainer easily. 

I think the explainer is still necessary when we want to launch this feature since we will be asked for a link to an explainer in many places. The explainer is also a good place to show why we want to add contextual biasing, and provides brief introduction on the changes for people who don't want to learn about every detail in specs.

padenot:
> I think the explainer is still necessary when we want to launch this feature since we will be asked for a link to an explainer in many places. The explainer is also a good place to show why we want to add contextual biasing, and provides brief introduction on the changes for people who don't want to learn about every detail in specs.

It's not necessary. Details can and should be in informative notes or MDN.

yrw-google:
> > I think the explainer is still necessary when we want to launch this feature since we will be asked for a link to an explainer in many places. The explainer is also a good place to show why we want to add contextual biasing, and provides brief introduction on the changes for people who don't want to learn about every detail in specs.
> 
> It's not necessary. Details can and should be in informative notes or MDN.

I've closed the PR for explainer. Can you now review the spec changes?

index.bs
```diff
@@ -162,12 +162,14 @@ interface SpeechRecognition : EventTarget {
...
     attribute unsigned long maxAlternatives;
     attribute SpeechRecognitionMode mode;
+    attribute SpeechRecognitionContext context;
 
     // methods to drive the speech interaction
     undefined start();
     undefined start(MediaStreamTrack audioTrack);
     undefined stop();
     undefined abort();
+    undefined updateContext(SpeechRecognitionContext context);
```
padenot, line 172: Method or attribute, is the above attribute readonly, is the method only used to change the attribute, etc.?

index.bs
```diff
@@ -322,12 +352,22 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
   The user agent must raise an <a event for=SpeechRecognition>end</a> event once the speech service is no longer connected.
   If the abort method is called on an object which is already stopped or aborting (that is, start was never called on it, the <a event for=SpeechRecognition>end</a> or <a event for=SpeechRecognition>error</a> event has fired on it, or abort was previously called on it), the user agent must ignore the call.</dd>
 
+  <dt><dfn method for=SpeechRecognition>updateContext({{SpeechRecognitionContext}} |context|)</dfn> method</dt>
+  <dd>
+    The updateContext method updates the speech recognition context after the speech recognition session has started.
+    If the session has not started yet, user should update {{SpeechRecognition/context}} instead of using this method.
```
padenot, line 358: Either/or. We don't need multiple ways to do a thing. What does update mean? Please use an algorithm with steps, that add or remove items from a list. The list is passed to an abstract object that is a speech recognition backend, that has its own state machine. It can likely only receive new context in a specific state. This needs to be stated and steps need to be unambiguous.

index.bs
```diff
@@ -322,12 +352,22 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+
+    When invoked, run the following steps:
+    1. If {{[[started]]}} is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
+    1. If the system does not support speech recognition context, throw a {{SpeechRecognitionErrorEvent}} with the {{context-not-supported}} error code and abort these steps.
+    1. The system updates its speech recognition context to be |context|.
```
padenot, line 363: Too hand-wavy. Does it replace the old information, is it an union, an intersection with the old information?

I would make the "context" an abstract concept that can be modified. One can add/remove items from it, each items has a boost, etc. Then you define `updateContext` properly by stating what it does with this abstract interface.

Or you just drop it altogether, and we just use the attribute.

E.g. you could do things like:

```js
let sr = new SpeechRecognition();
// ...
// ...
// Set some context items:
let phrase1 = new SpeechRecognitionPhrase("qqweqweqwe", 3.0);
let phrase2 = new SpeechRecognitionPhrase("asdasdasdas", 5.0);
let list = new SpeechRecognitionPhraseList(); // this should accept a sequence
list.addItem(phrase1);
list.addItem(phrase2);
sr.context = list;
let phrase3 = new SpeechRecognitionPhrase("iueriueureirue", 8.0);
list.addItem(phrase3);
sr.context = list;
```

what happens? Do have have multiple copies of `phrase1`, `phrase2`, or a single one. What happens if there are multiple copies of the same sentence, is there an order relationship, is it an error to do so?

index.bs
```diff
@@ -515,6 +561,66 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+  </dd>
+</dl>
+
+<h4 id="speechreco-phraselist">SpeechRecognitionPhraseList</h4>
+
+<p>The SpeechRecognitionPhraseList object holds a sequence of phrases for contextual biasing.</p>
+
+<dl>
+  <dt><dfn constructor for=SpeechRecognitionPhraseList>SpeechRecognitionPhraseList()</dfn> constructor</dt>
+  <dd>This constructor returns an empty list.</dd>
```
padenot, line 595: Make it accept a sequence for convenience.

index.bs
```diff
@@ -515,6 +561,66 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+
+  <dt><dfn attribute for=SpeechRecognitionPhraseList>length</dfn> attribute</dt>
+  <dd>This attribute indicates how many phrases are in the list. The user agent must ensure it is set to the number of phrases in the list.</dd>
```
padenot, line 598: Make this steps in an algorithm.

index.bs
```diff
@@ -515,6 +561,66 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+
+  <dt><dfn method for=SpeechRecognitionPhraseList>SpeechRecognitionPhrase(|index|)</dfn> method</dt>
+  <dd>
+    This method gets the SpeechRecognitionPhrase object at the |index| of the list.
+    When invoked, run the following steps:
+    1. If the |index| is smaller than 0, or greater than or equal to {{SpeechRecognitionPhraseList/length}}, return null.
```
padenot, line 604: Usually this throws `IndexSizeError`.

index.bs
```diff
@@ -515,6 +561,66 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+<p>The SpeechRecognitionPhraseList object holds a sequence of phrases for contextual biasing.</p>
+
+<dl>
+  <dt><dfn constructor for=SpeechRecognitionPhraseList>SpeechRecognitionPhraseList()</dfn> constructor</dt>
+  <dd>This constructor returns an empty list.</dd>
+
+  <dt><dfn attribute for=SpeechRecognitionPhraseList>length</dfn> attribute</dt>
+  <dd>This attribute indicates how many phrases are in the list. The user agent must ensure it is set to the number of phrases in the list.</dd>
+
+  <dt><dfn method for=SpeechRecognitionPhraseList>SpeechRecognitionPhrase(|index|)</dfn> method</dt>
```
padenot, line 600: This hasn't been updated to say `item`.

index.bs
```diff
@@ -515,6 +561,66 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+  <dd>
+    This method gets the SpeechRecognitionPhrase object at the |index| of the list.
+    When invoked, run the following steps:
+    1. If the |index| is smaller than 0, or greater than or equal to {{SpeechRecognitionPhraseList/length}}, return null.
+    1. Return the SpeechRecognitionPhrase from the |index| of the list.
+  </dd>
+
+  <dt><dfn method for=SpeechRecognitionPhraseList>addItem(|item|)</dfn> method</dt>
+  <dd>This method adds the SpeechRecognitionPhrase object |item| to the end of the list.</dd>
```
padenot, line 609: How can I remove an item.

index.bs
```diff
@@ -515,6 +561,66 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+<h4 id="speechreco-context">SpeechRecognitionContext</h4>
+
+<p>The SpeechRecognitionContext object holds contextual information to provide to the speech recognition models.</p>
+
+<dl>
+  <dt><dfn constructor for=SpeechRecognitionContext>SpeechRecognitionContext(|phrases|)</dfn> constructor</dt>
+  <dd>This constructor returns a new SpeechRecognitionContext object with the SpeechRecognitionPhraseList object |phrases| in it.</dd>
+
+  <dt><dfn attribute for=SpeechRecognitionContext>phrases</dfn> attribute</dt>
+  <dd>This attribute represents the phrases to be boosted.</dd>
```
padenot, line 621: Make this steps returning a private slot.

index.bs
```diff
@@ -515,6 +561,66 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+  <dd>This method adds the SpeechRecognitionPhrase object |item| to the end of the list.</dd>
+</dl>
+
+<h4 id="speechreco-context">SpeechRecognitionContext</h4>
+
+<p>The SpeechRecognitionContext object holds contextual information to provide to the speech recognition models.</p>
+
+<dl>
+  <dt><dfn constructor for=SpeechRecognitionContext>SpeechRecognitionContext(|phrases|)</dfn> constructor</dt>
+  <dd>This constructor returns a new SpeechRecognitionContext object with the SpeechRecognitionPhraseList object |phrases| in it.</dd>
```
padenot, line 618: This should just take a sequence.

index.bs
```diff
@@ -277,6 +304,9 @@ dictionary SpeechRecognitionEventInit : EventInit {
 
   <dt><dfn attribute for=SpeechRecognition>mode</dfn> attribute</dt>
   <dd>An enum to determine where speech recognition takes place. The default value is "ondevice-preferred".</dd>
+
+  <dt><dfn attribute for=SpeechRecognition>context</dfn> attribute</dt>
+  <dd>This attribute will set the speech recognition context for the recognition session to start with.</dd>
```
padenot, line 309: An attribute doesn't set anything. An attribute gets set. When this happens, a set of steps is triggered, described in an algorithm. For example here, we can decide this is readonly, and that any modification is to be performed using `updateContext`.

padenot (changes requested):
Getting there! lmk if I have been unclear in certain parts of the review.

index.bs
yrw-google: We will always set the attribute and I've removed the method.

index.bs
yrw-google: I've removed the updateContext method.

index.bs
yrw-google: We will just use the attribute. Setting the attribute will always replace it with the new value.

index.bs
```diff
@@ -515,6 +561,66 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+
+<p>The SpeechRecognitionPhrase object represents a phrase for contextual biasing.</p>
+
+<dl>
+  <dt><dfn constructor for=SpeechRecognitionPhrase>SpeechRecognitionPhrase(|phrase|, |boost|)</dfn> constructor</dt>
+  <dd>
+    When invoked, run the following steps:
+    1. If the |phrase| is an empty string, throw a "{{SyntaxError}}" {{DOMException}}.
+    1. If the |boost| is smaller than 0.0 or greater than 10.0, throw a "{{SyntaxError}}" {{DOMException}}.
+    1. Construct a new SpeechRecognitionPhrase object with |phrase| and |boost|.
```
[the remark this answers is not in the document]
yrw-google: I've updated according to your comments.

index.bs
yrw-google: Done.

index.bs
yrw-google: I've updated and the algorithm is just one line.

index.bs
yrw-google: Done.

index.bs
yrw-google: I've added a removeItem method.

index.bs
yrw-google: I've removed SpeechRecognitionContext so this no longer applies.

index.bs
yrw-google: Since SpeechRecognitionPhraseList already takes in a sequence, and SpeechRecognitionContext doesn't seem necessary at this point, I've removed SpeechRecognitionContext to avoid confusion and we can add SpeechRecognitionPhraseList to SpeechRecognition directly.

index.bs
yrw-google: I've added setter steps for this attribute.

yrw-google:
Hi @padenot, I decided to remove `SpeechRecognitionContext` and `updateContext()` since they seem redundant at this point and can cause confusion, so instead we will add `SpeechRecognitionPhraseList phrases` to SpeechRecognition directly and always update this attribute. I've also addressed your other comments so please take a look again.

index.bs
```diff
@@ -305,8 +297,14 @@ interface SpeechRecognitionContext {
...
 
-  <dt><dfn attribute for=SpeechRecognition>context</dfn> attribute</dt>
-  <dd>This attribute will set the speech recognition context for the recognition session to start with.</dd>
+  <dt><dfn attribute for=SpeechRecognition>phrases</dfn> attribute</dt>
+  <dd>
+    This attribute represents a list of phrases for contextual biasing.
+    The setter steps are:
+    1. If the {{SpeechRecognitionPhraseList/length}} of the given value is greater than 0 and the system does not support contextual biasing,
+        throw a {{SpeechRecognitionErrorEvent}} with the {{phrases-not-supported}} error code and abort these steps.
+    1. Set phrases to the given value.
```
padenot, line 306: We want what is called a "private slot" here. Think of it as essentially an attribute, on our instance.

We often use the notation `[[name]]`. Here, the second step would become:

```
1. Set [[phrases]] to the value set to this attribute.
```

index.bs
```diff
@@ -384,9 +371,10 @@ following steps:
 1. If |requestMicrophonePermission| is `true` and [=request
     permission to use=] "`microphone`" is [=permission/"denied"=], abort
     these steps.
-1. If {{SpeechRecognition/context}} is not null and the system does not support
-    speech recognition context, throw a {{SpeechRecognitionErrorEvent}} with the
-    {{context-not-supported}} error code and abort these steps.
+1. If the {{SpeechRecognitionPhraseList/length}} of {{SpeechRecognition/phrases}}
+    is greater than 0 and the system does not support contextual biasing, throw a
+    {{SpeechRecognitionErrorEvent}} with the {{phrases-not-supported}} error code
+    and abort these steps.
```
padenot, line 377: Can this happen? We have thrown when setting in those exact conditions, so this cannot happen, right?

index.bs
```diff
@@ -563,62 +551,91 @@ For a non-continuous recognition it will hold only a single value.</p>
 
 <h4 id="speechreco-phrase">SpeechRecognitionPhrase</h4>
 
-<p>The SpeechRecognitionPhrase object represents a phrase for contextual biasing.</p>
+<p>The SpeechRecognitionPhrase object represents a phrase for contextual biasing and has the following internal slots:</p>
+
+<dl dfn-type=attribute dfn-for="SpeechRecognitionPhrase">
+    : <dfn>[[phrase]]</dfn>
+    ::
+        A DOMString representing the text string to be boosted. The initial value is null.
```
padenot, line 559: Markup around `DOMString`.

index.bs
```diff
@@ -563,62 +551,91 @@ For a non-continuous recognition it will hold only a single value.</p>
...
 <dl>
   <dt><dfn constructor for=SpeechRecognitionPhrase>SpeechRecognitionPhrase(|phrase|, |boost|)</dfn> constructor</dt>
   <dd>
-    When invoked, run the following steps:
-    1. If the |phrase| is an empty string, throw a "{{SyntaxError}}" {{DOMException}}.
-    1. If the |boost| is smaller than 0.0 or greater than 10.0, throw a "{{SyntaxError}}" {{DOMException}}.
-    1. Construct a new SpeechRecognitionPhrase object with |phrase| and |boost|.
-    1. Return the object.
+    When this constructor is invoked, run the following steps:
+    1. If |phrase| is an empty string, throw a {{SyntaxError}} and abort these steps.
```
padenot, line 576: Why not ignore?

index.bs
```diff
@@ -563,62 +551,91 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+    1. If |boost| is smaller than 0.0 or greater than 10.0, throw a {{SyntaxError}} and abort these steps.
+    1. Let |phr| be a new object of type {{SpeechRecognitionPhrase}}.
+    1. Set |phr|.{{[[phrase]]}} to be |phrase|.
+    1. Set |phr|.{{[[boost]]}} to be |boost|.
+    1. Return the object |phr|.
```
padenot, line 581: We usually simply say: `Return |phr|.`

index.bs
```diff
@@ -563,62 +551,91 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+</dl>
 
 <dl>
-  <dt><dfn constructor for=SpeechRecognitionPhraseList>SpeechRecognitionPhraseList()</dfn> constructor</dt>
-  <dd>This constructor returns an empty list.</dd>
+  <dt><dfn constructor for=SpeechRecognitionPhraseList>SpeechRecognitionPhraseList(|phrases|)</dfn> constructor</dt>
+  <dd>
+    When this constructor is invoked, run the following steps:
+    1. Let |list| be a new object of type {{SpeechRecognitionPhraseList}}.
+    1. Set |list|.{{[[phrases]]}} to be |phrases|.
```
padenot, line 606:
```suggestion
    1. Set |list|.{{[[phrases]]}} to the value of |phrases|.
```

padenot: Here, are we making an internal copy of the list, or not? I think we are. If, further down the list, I modify `|phrases|`, I don't expect my changes to be reflected when transcribing. Or do I?

Are we setting a reference, or an actual list full of string+boost value? I would say the latter.

How do I update my biasing while recognizing? Should I keep my  `SpeechRecognitionPhraseList` around, remove some items, add some items, and set it again? That seems somewhat reasonnable, based on my experience (limited) experience with ASR systems. This was explicitely handled in the previous iteration of this PR.

index.bs
padenot: Additionally, we want to have some prose that "commits" the set of phrases in the ASR implementation, so that assigning to `phrases` is clearly the point at which the change happens internally. Something as vague as "send a copy of the phrase list to the implementation, etc.", will suffice, making it clear that (1) no method is required post-setting the attribute to "activate" those new phrases (2) modifying a reference to the list post-setting it has no effect, and re-setting the attribute is necessary. 

We need getter steps as well (they are to trivially return the value of `[[phrases]]`, e.g. as it often is the case, the setter has an effect, but the getter always ever returns what has been set).

padenot (changes requested):
Looking a lot better, some questions still, but it feels we're almost there.

index.bs
yrw-google: I've added `[[phrases]]` as a private slot of `SpeechRecognition`. I think we can still use the word [the given value](https://webidl.spec.whatwg.org/#the-given-value) to reference the input value. I also added a step of "Send a copy..." and the getter steps.

index.bs
yrw-google: I've removed this step. I was actually thinking about a case probably for Chrome only - Chrome supports contextual biasing only for on device speech recognition, but not for cloud speech recognition. If user sets [mode](https://webaudio.github.io/web-speech-api/#speechreco-attributes) to "ondevice-preferred", sets the phrase list, and then sets the mode to "cloud-only", we should return the `phrases-not-supported` error. This probably should be a step in the setter steps for [mode](https://webaudio.github.io/web-speech-api/#speechreco-attributes) instead of here, but since this probably only applies to Chrome, we can also just not mention it in the spec? Let me know if you think there's a better way to indicate this.

index.bs
yrw-google: Done.

index.bs
yrw-google: I think passing in an empty phrase with a boost value does not make sense to the speech recognition model, so we should tell the user that this should not happen. Although ignoring also works and the speech recognition model will just ignore this phrase. Let me know if you think there's an advantage of ignoring, otherwise I feel like we should prevent it if it's meaningless.

index.bs
yrw-google: Done.

index.bs
yrw-google: We are always making a copy of the text+boost list rather than reference. If user wants to update the phrases, they need to set the `phrases` in the SpeechRecognition attributes again with the updated `SpeechRecognitionPhraseList`. I've indicated this by adding the step "Send a copy..." in the setter steps for `phrases`.

index.bs
padenot: Firefox will only support on-device recognition, so it does not apply to us, but if you think we need this in some cases, then it's best to keep it. We'll simply not implement this on our side, but Chrome will have a legal behaviour.

I was thinking solely in terms of local recognition when reviewing this, and completely forgot it was only part of the system.

yrw-google: I've added an internal slot for `mode` and setter steps in `mode` to indicate this. I'm not sure if the setter steps for `mode` needs to involve more steps not related to biasing, but we can discuss with @evanbliu later if you have concerns. Please take a look at the changes again and let me know if you have other questions.

yrw-google:
Hi @padenot, a gentle ping for reviewing this

index.bs
```diff
@@ -276,7 +307,34 @@ dictionary SpeechRecognitionEventInit : EventInit {
...
+   This attribute represents where speech recognition takes place.
+  </dd>
+  <dd>
+    The getter steps are to return the value of {{SpeechRecognition/[[mode]]}}.
+  </dd>
+  <dd>
+    The setter steps are:
+    1. If the {{SpeechRecognitionPhraseList/length}} of {{SpeechRecognition/phrases}} is greater than 0
+        and the system using the given value for {{SpeechRecognition/[[mode]]}} does not support contextual biasing,
+        throw a {{SpeechRecognitionErrorEvent}} with the {{SpeechRecognitionErrorCode/not-allowed}} error code and abort these steps.
```
padenot, line 320: `not-supported` for consistency? Or we want to clearly distinguish the two? What do we lose by using "normal" errors here?

index.bs
```diff
@@ -515,6 +575,95 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+</dl>
+
+<h4 id="speechreco-phraselist">SpeechRecognitionPhraseList</h4>
+
+<p>The SpeechRecognitionPhraseList object holds a sequence of phrases for contextual biasing and has the following internal slot:</p>
+
+<dl dfn-type=attribute dfn-for="SpeechRecognitionPhraseList">
+    : <dfn>[[phrases]]</dfn>
+    ::
+        A sequence of {{SpeechRecognitionPhrase}} representing the phrases to be boosted. The initial value is an empty list.
```
padenot, line 624: When "inside" the binding, we don't use concepts such as WebIDL's `sequence<...>`, but rather things from https://infra.spec.whatwg.org/#lists, here a list.

This means "removing" and "appending" is well-defined, and there is no question or doubt about shifting elements that are after an element that has just been removed on position to the left, and all that.

index.bs
```diff
@@ -515,6 +575,95 @@ For a non-continuous recognition it will hold only a single value.</p>
...
+    When invoked, run the following steps:
+    1. If |index| is smaller than 0, or greater than or equal to {{SpeechRecognitionPhraseList/length}},
+        throw a {{RangeError}} and abort these steps.
+    1. Return the {{SpeechRecognitionPhrase}} at the |index| of {{SpeechRecognitionPhraseList/[[phrases]]}}.
+  </dd>
+
+  <dt><dfn method for=SpeechRecognitionPhraseList>addItem(|item|)</dfn> method</dt>
+  <dd>
+    This method adds the {{SpeechRecognitionPhrase}} object |item| to the list.
+    When invoked, add |item| to the end of {{SpeechRecognitionPhraseList/[[phrases]]}}.
```
padenot, line 654: What happens if we add an element twice with different scores? Do we want to use a set instead from the `infra` link above?

index.bs
padenot: Sometimes we like to be a bit lenient. Consider the following bit of code, that takes the description of the video in some sort of metadata to bias the recognition:

```
let phrase_list = new SpeechRecognitionPhraseList();
let sentences = videoMetadata.description.split(".");
let regular_list = [];
sentences.forEach((s) => {
  let phrase = new SpeechRecognitionPhrase(s, 5.0);
  phrase_list.append(phrase);
  regular_list.push(phrase);
});

// or just use the sequence ctor
let phrase_list_2 = new SpeechRecognitionPhraseList(regular_list);
```

This is all well and good. Now what if the description is:

```
"In this episode, our heroes will encounter a terrible revelation...\nWill they triumph?"
```

This will have empty elements, that will make the above code throw. It's a trivial fix, but it's a fix that all web developers will have to write. Same for string that are only whitespace, etc. https://infra.spec.whatwg.org/#ascii-whitespace helpfully exists if we want to handle this case (empty or whitespace-only, etc.).

padenot (changes requested):
Almost there!

index.bs
yrw-google: Ok, I'll remove the check for the phrase.

index.bs
yrw-google: I was thinking that when user is trying to set the mode here but get a `phrases-not-supported` error, it seems weird that the error is not related to mode directly, but we can also add error message to explain this better, so I'll switch to `phrases-not-supported` for consistency.

index.bs
yrw-google: I'll use list instead.

index.bs
yrw-google: If we want to avoid the same phrase with different boosts, we will need to use a map like `map<string, float>` for the phrase-boost pair, rather than a set. I feel like this adds more limitations to the design which is not as desired. For example, if in the future we want to add a pronunciation attribute for each `SpeechRecognitionPhrase`, using a list like now is easier than a map.

For now if user adds the phrase twice with different boosts, I think the system can use the boost that's updated at the second phrase, and this case should be rare.

padenot: Right, my point is that we can do whatever, but it needs to be specified.

yrw-google: I've added descriptions for this case.

yrw-google:
Hi @padenot, another ping on reviewing this.

padenot (approved):

yrw-google:
Hi @padenot, thanks for approving but I didn't find how I can merge this pull request. Since it shows "1 workflow awaiting approval" and "This workflow requires approval from a maintainer", does it mean you need to run something on your side to start the merge?

padenot:
Thanks @evanbliu for sorting it out.

PR #146, opened 2025-02-28
Add Makefile to replicate CI

padenot:


Makefile
```diff
@@ -0,0 +1,33 @@
...
+	@ echo "All done"
+
+%.html : %.bs node_modules/vnu-jar/build/dist/vnu.jar
+ifndef REMOTE
+	@ echo "Building $@"
+	bikeshed --die-on=warning spec $< $@
+	java -jar node_modules/vnu-jar/build/dist/vnu.jar --also-check-css $@
+else
+	@ echo "Building $@ remotely"
+	@ (HTTP_STATUS=$$(curl https://api.csswg.org/bikeshed/ \
```
hoch, line 16: This is interesting! I wish we could do the same with the Web Audio API.

Makefile
```diff
@@ -0,0 +1,33 @@
...
+REMOTE := $(filter remote,$(MAKECMDGOALS))
+
+all: $(DST)
+	@ echo "All done"
+
+%.html : %.bs node_modules/vnu-jar/build/dist/vnu.jar
+ifndef REMOTE
+	@ echo "Building $@"
+	bikeshed --die-on=warning spec $< $@
+	java -jar node_modules/vnu-jar/build/dist/vnu.jar --also-check-css $@
```
hoch, line 13: Is this new dependency on `vnu.jar` necessary?

hoch (approved):
LGTM, but I have one question.

Makefile
padenot: Our CI checks it, so I figured we could make it easy to run locally.

Here's an example of a failure in a previous PR of mine: https://github.com/WebAudio/web-speech-api/actions/runs/13461066795?pr=144 (which was legitimate, my markup was incorrect).

PR #157, opened 2025-05-07
Update spec with latest on-device changes

### LINKED ISSUE #155: Consider changing installOnDevice() parameter to InstallOptions
Currently the installOnDevice API takes in a collection of languages. We should consider future-proofing this API by encapsulating the collection of languages in an InstallOptions object. That way in the future we have the flexibility of extending this API to allow websites to install on-device speech recognition with specific capabilities without breaking backwards compatibility.

### LINKED ISSUE #156: Consider renaming SpeechRecognitionMode enum values
From the TAG feedback:

> We do think there's a reasonable use case in ensuring that audio data isn't spread around more widely than it needs to be, but "ondevice" doesn't seem to capture the right distinction. In [discussion](https://github.com/w3ctag/meetings/blob/gh-pages/2025/telcons/04-28-minutes.md#design-reviews1038-on-device-web-speech-api---jyasskin), we identified 5 levels of exposure, and there may be more:
> 
> A. The same device that the UA is running on.
> B. Encrypted connections to other devices that the same user owns, for example to allow a phone to offload work to the same person's desktop.
> C. Various flavors of private computation on other people's computers, ranging from homomorphic encryption to [TEEs](https://en.wikipedia.org/wiki/Trusted_execution_environment).
> D. "Service providers" (in the GDPR sense) for the user, who have committed not to use or expose the data outside that particular user's context.
> E. Distinct data controllers (again in the GDPR sense), who could use the user's data for their own purposes. This secondary use might help subsidize free or cheaper service, but it also makes this category challenging.
> 
> "ondevice" implies at most level A, but the E2EE and regulatory use cases could handle at least level B and maybe up to level D. We'd like the WG to be more explicit about what's needed here, and to pick a term that captures the right distinctions. 

More context available here: https://github.com/w3ctag/design-reviews/issues/1038#issuecomment-2803693504

evanbliu:
Closes #155
Closes #156

This PR contains the following changes:

- Rename availableOnDevice() to available() and installOnDevice() to install().
- Update available() and install() parameters to take in a dictionary encapsulating the on-device model options.
- Add a note on forbidding personalization when recognizing audio from MediaStreamTracks.
- Remove SpeechRecognitionMode and replace it with SpeechRecognitionOptions.processLocally.
- Update the install() and available() methods to handle a collection of languages.

***
<a href="https://pr-preview.s3.amazonaws.com/evanbliu/speech-api/pull/157.html" title="Last updated on May 27, 2025, 8:39 PM UTC (0a1b233)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/157/af3de99...evanbliu:0a1b233.html" title="Last updated on May 27, 2025, 8:39 PM UTC (0a1b233)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -119,7 +119,7 @@ This does not preclude adding support for this as a future API enhancement, and 
 
   <li>The user agent may also give the user a longer explanation the first time speech input is used, to let the user know what it is and how they can tune their privacy settings to disable speech recording if required.</li>
 
-  <li>To mitigate the risk of fingerprinting, user agent MUST NOT personalize speech recognition when performing speech recognition on a {{MediaStreamTrack}}.</li>
+  <li>To mitigate the risk of fingerprinting, user agents MUST NOT personalize speech recognition when performing speech recognition on a {{MediaStreamTrack}}.</li>
 </ol>
 
 <h3 id="implementation-considerations">Implementation considerations</h3>
@@ -143,8 +143,8 @@ This does not preclude adding support for this as a future API enhancement, and 
 <h3 id="speechreco-section">The SpeechRecognition Interface</h3>
 
 <p>The speech recognition interface is the scripted web API for controlling a given recognition.</p>
-The term "final result" indicates a SpeechRecognitionResult in which the final attribute is true.
-The term "interim result" indicates a SpeechRecognitionResult in which the final attribute is false.
+The term "final result" indicates a {{SpeechRecognitionResult}} in which the {{SpeechRecognitionResult/isFinal}} attribute is true.
+The term "interim result" indicates a {{SpeechRecognitionResult}} in which the {{SpeechRecognitionResult/isFinal}} attribute is false.
 
 {{SpeechRecognition}} has the following internal slots:
 
@@ -319,22 +319,16 @@ interface SpeechRecognitionPhraseList {
   <dd>
    This attribute allows specification of a {{SpeechRecognitionOptions}} object to configure options for speech recognition.
    If this attribute is null, default behaviors apply for these options.
-   The {{SpeechRecognitionOptions}} dictionary has the following members:
+   Setting this attribute sets internal slots for the {{SpeechRecognitionOptions}} dictionary which has the following members:
    <dl>
      <dt><dfn dict-member for="SpeechRecognitionOptions">langs</dfn></dt>
      <dd>A sequence of {{DOMString}}s. Each {{DOMString}} in the sequence <em class="rfc2119" title="MUST">MUST</em> be a
      valid [[!BCP47]] language tag. When used with {{SpeechRecognition/start()}}, the user agent will attempt to use the languages in the order provided, selecting the first one that is 
-     available. When used with {{SpeechRecognition/start()}} and {{SpeechRecognition/lang}} is also specified, {{SpeechRecognitionOptions/langs}} <em class="rfc2119" title="MUST">MUST</em> be empty or contain a single language tag identical to {{SpeechRecognition/lang}}.</dd>
+     available. 
      <dt><dfn dict-member for="SpeechRecognitionOptions">processLocally</dfn></dt>
      <dd>A boolean that defaults to <code>false</code>.
-     If set to <code>true</code>, it indicates a requirement that the speech recognition process <em class="rfc2119" title="MUST">MUST</em> be performed locally on the user's device, without sending audio data to a remote server.
-     If set to <code>false</code>, the user agent is free to choose between local and remote processing.
-     <p>When the {{SpeechRecognition/options}} attribute of a {{SpeechRecognition}} instance is set to a {{SpeechRecognitionOptions}} object where {{SpeechRecognitionOptions/processLocally}} is <code>true</code>,
-     and {{SpeechRecognition/start()}} is subsequently called: if local processing is not available for the language selected or if the user agent cannot fulfill the local processing requirement for other reasons,
-     the {{SpeechRecognition/start()}} method <em class="rfc2119" title="MUST">MUST</em> fail by dispatching an <a event for=SpeechRecognition>error</a> event with the {{SpeechRecognitionErrorCode/service-not-allowed}} code.</p>
-     <p>When used with {{SpeechRecognition/available()}}, it will check the availability of speech recognition matching the options.
-     When used with {{SpeechRecognition/install()}} it will install speech recognition matching the options.</p>
-     </dd>
+     If set to <code>true</code>, it indicates a requirement that the speech recognition process <em class="rfc2119" title="MUST">MUST</em> be performed locally on the user's device.
+     If set to <code>false</code>, the user agent can choose between local and remote processing.</dd>
    </dl>
   </dd>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -199,6 +199,14 @@ interface SpeechRecognition : EventTarget {
     attribute EventHandler onend;
 };
 
+dictionary SpeechRecognitionAvailabilityOptions {
```
evanbliu, line 202: What are your thoughts on having separate dictionaries for availability/installation? These two are currently identical, but they allow the availability options to differ from the installation options in the future.

padenot: Can you give an example of what those would be?

Besides the language, I can see the following:
- quality / speed type of thing
- diarization capability

evanbliu: We could also rename availableOnDevice() and installOnDevice() to available() and install() and include the on-device part (or the options described here: https://github.com/WebAudio/web-speech-api/issues/156#issuecomment-2860911476) in the SpeechRecognitionOptions.

padenot: We could yes.

evanbliu:
@padenot - I updated this PR with the rest of the changes we discussed at the Audio WG meeting yesterday. Please take a look when you get a chance :)

index.bs
```diff
@@ -314,20 +315,27 @@ interface SpeechRecognitionPhraseList {
...
+   If this attribute is null, default behaviors apply for these options.
+   The {{SpeechRecognitionOptions}} dictionary has the following members:
+   <dl>
+     <dt><dfn dict-member for="SpeechRecognitionOptions">langs</dfn></dt>
+     <dd>A sequence of {{DOMString}}s. Each {{DOMString}} in the sequence <em class="rfc2119" title="MUST">MUST</em> be a
+     valid [[!BCP47]] language tag. When used with {{SpeechRecognition/start()}}, the user agent will attempt to use the languages in the order provided, selecting the first one that is 
+     available. When used with {{SpeechRecognition/start()}} and {{SpeechRecognition/lang}} is also specified, {{SpeechRecognitionOptions/langs}} <em class="rfc2119" title="MUST">MUST</em> be empty or contain a single language tag identical to {{SpeechRecognition/lang}}.</dd>
+     <dt><dfn dict-member for="SpeechRecognitionOptions">processLocally</dfn></dt>
+     <dd>A boolean that defaults to <code>false</code>.
+     If set to <code>true</code>, it indicates a requirement that the speech recognition process <em class="rfc2119" title="MUST">MUST</em> be performed locally on the user's device, without sending audio data to a remote server.
```
padenot, line 330: I think it would be cleaner to flip the sentence here, not saying that it should not be sent out, but saying it should stay on the device. Here, I can clearly see someone nitpicking and saying it's legal to send the audio to something that isn't a server, which isn't our intent.

index.bs
```diff
@@ -314,20 +315,27 @@ interface SpeechRecognitionPhraseList {
...
+     If set to <code>false</code>, the user agent is free to choose between local and remote processing.
```
padenot, line 331: `s/is free/can/` (more usual in specs).

index.bs
```diff
@@ -314,20 +315,27 @@ interface SpeechRecognitionPhraseList {
...
+     <p>When the {{SpeechRecognition/options}} attribute of a {{SpeechRecognition}} instance is set to a {{SpeechRecognitionOptions}} object where <code>processLocally</code> is <code>true</code>,
```
padenot, line 332:
```suggestion
     <p>When the {{SpeechRecognition/options}} attribute of a {{SpeechRecognition}} instance is set to a {{SpeechRecognitionOptions}} object where {{SpeechRecognitionOptions/processLocally}} is <code>true</code>,
```

index.bs
```diff
@@ -314,20 +315,27 @@ interface SpeechRecognitionPhraseList {
...
+     and {{SpeechRecognition/start()}} is subsequently called: if local processing is not available for the language selected or if the user agent cannot fulfill the local processing requirement for other reasons,
+     the {{SpeechRecognition/start()}} method <em class="rfc2119" title="MUST">MUST</em> fail by dispatching an <a event for=SpeechRecognition>error</a> event with the {{SpeechRecognitionErrorCode/service-not-allowed}} code.</p>
```
padenot, line 334: In general it's nicer to put this bit of behaviour in the start method.

We should change this prose to use the style we use elsewhere: setting this attribute sets internal slots (`[[language]]`, `[[process locally]]`). In `start()`, if `[[processLocally]]` is `true`, but the implementation cannot process locally, then an exception is thrown.

Similarly for `available()` and `install()`, let's define what the value of this object mean there.

index.bs
```diff
@@ -389,46 +397,51 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
     1. Return <var>promise</var>.
   </dd>
 
-  <dt><dfn method for=SpeechRecognition>installOnDevice({{DOMString}} lang)</dfn> method</dt>
+  <dt><dfn method for=SpeechRecognition>install({{SpeechRecognitionOptions}} options)</dfn> method</dt>
   <dd>
-    The {{SpeechRecognition/installOnDevice}} method returns a {{Promise}} that resolves to a {{boolean}} when and whether the installation of on-device speech recognition for a given [[!BCP47]] language tag succeeded.
+    The {{SpeechRecognition/install}} method attempts to install on-device speech recognition language packs for all languages specified in `options.langs`.
+    It returns a {{Promise}} that resolves to a {{boolean}}.
+    The promise resolves to `true` if all installation attempts for requested and supported languages succeed (or the languages were already installed).
```
padenot, line 415:
```suggestion
    The promise resolves to `true` when all installation attempts for requested and supported languages succeed (or the languages were already installed).
```

padenot: Again we could write this as an algorithm.

index.bs
```diff
@@ -389,46 +397,51 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+    The promise resolves to `false` if `options.langs` is empty, if not all of the requested languages are supported, or if any installation attempt for a supported language fails.
```
padenot, line 416: Let's add a note that say that `processLocally` isn't used in this algorithm. This is non-normative, but clearly indicates it's not an oversight on our part.

index.bs
```diff
@@ -389,46 +397,51 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+    1. If the on-device speech recognition language pack for any <var>lang</var> in <code>options.langs</code> is unsupported, return a resolved {{Promise}} with false and skip the rest of these steps.
     1. Let <var>promise</var> be <a>a new promise</a>.
-    1. Initiate the download of the on-device speech recognition language for <var>lang</var>.
+    1. For each <var>lang</var> in <code>options.langs</code>, initiate the download of the on-device speech recognition language for <var>lang</var>.
         <p class=note>
           Note: The user agent can prompt the user for explicit permission to download the on-device speech recognition language pack.
         </p>
     1. [=Queue a task=] on the [=relevant global object=]'s [=task queue=] to run the following step:
-        - If the download succeeds, resolve <var>promise</var> with <code>true</code>, otherwise resolve it with <code>false</code>.
+        - If the download of all languages specified by <code>options.langs</code> succeeds, resolve <var>promise</var> with <code>true</code>, otherwise resolve it with <code>false</code>.
```
padenot, line 429:
```suggestion
        - When the download of all languages specified by <code>options.langs</code> succeeds, resolve <var>promise</var> with <code>true</code>, otherwise resolve it with <code>false</code>.
```

index.bs
```diff
@@ -389,46 +397,51 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
-1. If <var>lang</var> is not a valid [[!BCP47]] language tag, throw a {{SyntaxError}} and abort these steps.
-1. Determine the availability status for <var>lang</var>:
-    - If the on-device speech recognition language pack for <var>lang</var> is unsupported, let <var>status</var> be {{AvailabilityStatus/unavailable}}.
-    - Else if the on-device speech recognition language pack for <var>lang</var> is supported but not installed, let <var>status</var> be {{AvailabilityStatus/downloadable}}.
-    - Else if the on-device speech recognition language pack for <var>lang</var> is downloading, let <var>status</var> be {{AvailabilityStatus/downloading}}.
-    - Else if the on-device speech recognition language pack for <var>lang</var> is installed, let <var>status</var> be {{AvailabilityStatus/available}}.
+1. If any <var>lang</var> in <code>langs</code> is not a valid [[!BCP47]] language tag, throw a {{SyntaxError}} and abort these steps.
+1. Determine the availability status for <var>langs</var>:
+    - If on-device speech recognition for each <var>lang</var> in <code>langs</code> is installed, let <var>status</var> be {{AvailabilityStatus/available}}.
```
padenot, line 441:
```suggestion
    - If on-device speech recognition for all <var>lang</var> in <code>langs</code> is installed, let <var>status</var> be {{AvailabilityStatus/available}}.
```

index.bs
```diff
@@ -389,46 +397,51 @@ See <a href="https://lists.w3.org/Archives/Public/public-speech-api/2012Sep/0072
...
+    - Else if on-device speech recognition for each <var>lang</var> in <code>langs</code> is installed or downloading, let <var>status</var> be {{AvailabilityStatus/downloading}}.
+    - Else if on-device speech recognition for each <var>lang</var> in <code>langs</code> is installed, downloading, or supported but not installed let <var>status</var> be {{AvailabilityStatus/downloadable}}.
+    - Else let <var>status</var> be {{AvailabilityStatus/unavailable}}.
```
padenot, line 444: I think we can simplify this a bit, it's a bit unclear how language take precedence when more than one is stipulated.

Let's say that enum values are ordered: `available` < `downloading` < `downloadable` < `unavailable`, and say this returns `max(availability(lang))`. Or we can return an array?

evanbliu:
According to the [Web IDL spec](https://webidl.spec.whatwg.org/#idl-dictionaries), dictionaries can't be attributes, so I removed the options attribute from the SpeechRecognition interface and instead just added a boolean attribute for processLocally. The added benefit of this change is that there can no longer be a state where SpeechRecognition.lang and SpeechRecognition.options.langs have different values.

index.bs
```diff
@@ -153,9 +155,9 @@ The term "interim result" indicates a SpeechRecognitionResult in which the final
 </dl>
 
 <dl dfn-type=attribute dfn-for="SpeechRecognition">
-    : <dfn>[[mode]]</dfn>
+    : <dfn>[[processLocally]]</dfn>
     ::
-        A {{SpeechRecognitionMode}} enum to determine where speech recognition takes place. The initial value is <code>ondevice-preferred</code>.
+        A boolean flag indicating whether recognition must be performed locally. The initial value is <code>false</code>.
```
padenot, line 160: must all caps + markup

index.bs
```diff
@@ -439,9 +477,13 @@ following steps:
 1. If the [=current settings object=]'s [=relevant global object=]'s
      [=associated Document=] is NOT [=fully active=], throw an {{InvalidStateError}}
      and abort these steps.
-1. If {{[[started]]}} is `true` and no <a event
-     for=SpeechRecognition>error</a> or <a event for=SpeechRecognition>end</a> event
-     have fired, throw an {{InvalidStateError}} and abort these steps.
+1. If {{SpeechRecognition/[[started]]}} is `true` and no <a event
+    for=SpeechRecognition>error</a> event or <a event for=SpeechRecognition>end</a> event
+    has fired on it, throw an {{InvalidStateError}} and abort these steps.
```
padenot, line 482: We want to track this in a less hand-wavy way (flipping a bool when at least one has fired, and checking that).

padenot (approved):

PR #165, opened 2025-06-17
Add privacy and security mitigations to the Web Speech API

### LINKED ISSUE #158: Apply privacy and security mitigations used by the Web Translation API to the Web Speech API
Chrome is planning on implementing the same privacy and security measures used by the Web Translation API, which uses on-device language packs for translations, to the Web Speech API.

We should consider adding these mitigations to the Web Speech API spec as well: https://github.com/webmachinelearning/writing-assistance-apis/pull/47

TL;DR: Websites must call install() before they can use the on-device speech recognition language packs to prevent fingerprinting.

evanbliu:
This PR adds the privacy and security mitigations used by the [Writing Assistance APIs](https://webmachinelearning.github.io/writing-assistance-apis) to the on-device speech recognition part of the Web Speech API.

Closes #158

***
<a href="https://pr-preview.s3.amazonaws.com/evanbliu/speech-api/pull/165.html" title="Last updated on Mar 5, 2026, 12:15 AM UTC (99dd47a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/165/59b1992...evanbliu:99dd47a.html" title="Last updated on Mar 5, 2026, 12:15 AM UTC (99dd47a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -136,6 +136,116 @@ This does not preclude adding support for this as a future API enhancement, and 
   The above recommendations are intended to reduce this risk of such attacks.</li>
 </ol>
 
+<h3 id=on-device-model-privacy>On-Device Model Privacy Considerations</h3>
+<em>This subsection, and the "On-Device Model Security Considerations" subsection below, unlike many "privacy considerations" sections which only summarize and restate considerations that are already normatively specified elsewhere in the document, contain some normative requirements that are not present elsewhere, and add more detail to the normative requirements present elsewhere. The novel normative requirements are called out using <strong>strong emphasis</strong>.</em>
+
+<h4 id="on-device-model-privacy-availability">Language Pack Availability</h4>
+
+For on-device speech recognition, the exact download status of language packs can present a fingerprinting vector. How many bits this vector provides depends on the options provided to {{SpeechRecognition/available()}} or {{SpeechRecognition/install()}}, and how they influence the download (e.g., if different language packs have different availability statuses).
+
+<h4 id="on-device-model-privacy-availability-masking">Download Masking</h4>
+
+One mitigation is for the user agent to mask the current download status by returning {{"downloadable"}} from {{SpeechRecognition/available()}} even if the actual download status is {{"available"}} or {{"downloading"}}.
+
+Because implementation strategies differ and other mitigations (like permission prompts for {{SpeechRecognition/install()}}) are available, a specific masking scheme is not mandated. For APIs where the user agent believes such masking is necessary, a suggested heuristic is to mask by default, subject to a masking state that is established for each (API, options, [=storage key=]) tuple. This state can be set to "unmasked" once a web page in a given [=storage key=] calls {{SpeechRecognition/install()}} with a given set of options, and successfully starts a download or the promise resolves to `true` (indicating the language pack is ready). Since {{SpeechRecognition/install()}} has stronger requirements (see <a href="#on-device-model-privacy-availability-installation">Installation-time friction</a>), this ensures that web pages only get access to the true download status after taking a more costly and less-repeatable action.
+
+<strong>Implementations which use such a [=storage key=]-based masking scheme must ensure that the masking state is reset when other storage for that origin is reset.</strong>
+
+<h5 id="on-device-model-privacy-availability-installation">Installation-time friction</h5>
+
+The mitigation described in <a href="#on-device-model-privacy-availability-masking">Download Masking</a> works against attempts to silently fingerprint using {{SpeechRecognition/available()}}. The specification also contains requirements to prevent {{SpeechRecognition/install()}} from being easily used for fingerprinting, by introducing friction:
+
+*   The {{SpeechRecognition/install()}} method both requires and consumes [=user activation=], when it would initiate a download.
+*   The {{SpeechRecognition/install()}} method allows the user agent to prompt the user for permission, or to implicitly reject download attempts based on previous signals (such as an observed pattern of abuse).
+*   Access to {{SpeechRecognition/install()}} and {{SpeechRecognition/available()}} is gated on an per-API [=policy-controlled feature=], which means that only top-level origins and their delegates can use the API.
+
+Additionally, initiating the download process via {{SpeechRecognition/install()}} is more or less a one-time operation for a given language. The availability status will only transition from {{"downloadable"}} to {{"downloading"}} to {{"available"}} via these guarded installation operations. That is, while {{SpeechRecognition/install()}} can be used to read some of these fingerprinting bits (by observing the resolution of its promise and subsequent calls to {{SpeechRecognition/available()}}), doing so will effectively "destroy" those bits by changing the state.
+
+(For details on cases where downloading might happen more than once, and how privacy and security are preserved in those cases, see <a href="#on-device-model-privacy-availability-cancelation">Download Cancelation</a>, <a href="#on-device-model-privacy-availability-eviction">Download Eviction</a>, and <a href="#on-device-model-security-disk-space">Disk Space for Language Packs</a>.)
+
+<h5 id="on-device-model-privacy-availability-cancelation">Download Cancelation</h5>
+
+An important part of making the download status a less-useful fingerprinting vector is to ensure that the website cannot toggle the availability state back and forth by starting and then effectively canceling downloads. The Web Speech API's {{SpeechRecognition/install()}} method returns a promise and does not take an {{AbortSignal}} to cancel the download itself ({{SpeechRecognition/abort()}} is for an active recognition session).
+
+Once a download is initiated by {{SpeechRecognition/install()}}, the user agent should preserve the download progress. <strong>User agents should not cancel an ongoing language pack download in response to page-controlled actions</strong> (e.g., navigation, page unload) that could be used to manipulate the download state for fingerprinting. If a page navigates away, the download should ideally continue in the background, or at least its progress should be saved. The goal is to prevent the site from easily reverting a language pack's state from {{"downloading"}} back to {{"downloadable"}}.
+
+Note that canceling downloads in response to explicit, out-of-band user-controlled actions (e.g., via browser UI) is not problematic from this perspective.
+
+<h5 id="on-device-model-privacy-availability-eviction">Download Eviction</h5>
+
+Another ingredient in ensuring that websites cannot toggle the availability state back and forth is to ensure that user agents don't use a quota-based eviction system for downloaded language packs that web pages can indirectly control. For example, if a user agent evicted less-recently-used language packs when new ones are installed, a web page could trigger such evictions to toggle the state of a target language pack.
+
+To avoid this, <strong>user agents should not implement systems which allow web pages to control the eviction of downloaded language packs</strong>, including via indirect triggers such as further subsequent downloads. One way to fulfill this requirement is to never evict downloaded material in response to web page-initiated storage pressure, instead refusing to download new material (e.g., {{SpeechRecognition/install()}} resolving to `false`) if doing so would cause storage pressure.
+
+Evicting downloads in response to user-controlled actions (e.g., via a browser settings UI) is not problematic.
+
+<h5 id="on-device-model-privacy-availability-alternatives">Alternate Options</h5>
+
+While some of the above requirements are specified using "must" language, most are "should." This is because implementations might use different strategies to preserve user privacy, especially for APIs with smaller models or language packs.
+
+The simplest is to treat language pack downloads like other stored resources, partitioning them by the downloading page's [=storage key=]. This leverages existing web origin model privacy protections. The downside is potentially redundant downloads across sites, using more user bandwidth and disk space.
+
+A variant is to re-download for new [=storage keys=] but re-use on-disk storage if the pack is already there, saving disk space but still using time/bandwidth.
+
+User agents could also attempt to fake a download for new [=storage keys=] if the language pack is already present, by waiting a similar amount of time as the real download originally took. This saves bandwidth and disk space but is less private due to network side channels (e.g., a page observing no change in network throughput). Such a scheme needs caution, as the first site initiating the download could try to inflate this time. Nevertheless, faking download times might be useful, combined with other mitigations.
+
+<h4 id="on-device-model-privacy-sensitive-language">Sensitive Language Information</h4>
+
+Even if fingerprinting risks from availability status are mitigated, knowing a user has downloaded a specific language pack (e.g., for a minority language) can be sensitive.
+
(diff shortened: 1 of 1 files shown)


evanbliu:
@padenot - I just realized this PR was never merged. Can you please take a look when you get a chance?

index.bs
```diff
@@ -136,6 +139,116 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+<h4 id="on-device-model-privacy-availability">Language Pack Availability</h4>
+
+For on-device speech recognition, the exact download status of language packs can present a fingerprinting vector. How many bits this vector provides depends on the options provided to {{SpeechRecognition/available()}} or {{SpeechRecognition/install()}}, and how they influence the download (e.g., if different language packs have different availability statuses).
+
+<h4 id="on-device-model-privacy-availability-masking">Download Masking</h4>
+
+One mitigation is for the user agent to mask the current download status by returning {{"downloadable"}} from {{SpeechRecognition/available()}} even if the actual download status is {{"available"}} or {{"downloading"}}.
```
marcoscaceres: Shouldn't available be only available if and only if the site first requested it (i.e., it's origin bound)... that is, the site must risk downloading it (and potentially throwing up an install prompt) to add a language. 

And the availability must only be subject to that origin binding, and never reveal if another origin downloaded the same language. 

That should prevent fingerprinting - but also prevents this from being abused by multiple coordinating sites.

index.bs
```diff
@@ -136,6 +139,116 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+Because implementation strategies differ and other mitigations (like permission prompts for {{SpeechRecognition/install()}}) are available, a specific masking scheme is not mandated. For APIs where the user agent believes such masking is necessary, a suggested heuristic is to mask by default, subject to a masking state that is established for each (API, options, [=storage key=]) tuple. This state can be set to "unmasked" once a web page in a given [=storage key=] calls {{SpeechRecognition/install()}} with a given set of options, and successfully starts a download or the promise resolves to `true` (indicating the language pack is ready). Since {{SpeechRecognition/install()}} has stronger requirements (see <a href="#on-device-model-privacy-availability-installation">Installation-time friction</a>), this ensures that web pages only get access to the true download status after taking a more costly and less-repeatable action.
```
marcoscaceres: > "a specific masking scheme is not mandated"

I don't think this is good enough. See you if you can come up with a fix for this.

index.bs
```diff
@@ -136,6 +139,116 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+(For details on cases where downloading might happen more than once, and how privacy and security are preserved in those cases, see <a href="#on-device-model-privacy-availability-cancelation">Download Cancelation</a>, <a href="#on-device-model-privacy-availability-eviction">Download Eviction</a>, and <a href="#on-device-model-security-disk-space">Disk Space for Language Packs</a>.)
+
+<h5 id="on-device-model-privacy-availability-cancelation">Download Cancelation</h5>
+
+An important part of making the download status a less-useful fingerprinting vector is to ensure that the website cannot toggle the availability state back and forth by starting and then effectively canceling downloads. The Web Speech API's {{SpeechRecognition/install()}} method returns a promise and does not take an {{AbortSignal}} to cancel the download itself ({{SpeechRecognition/abort()}} is for an active recognition session).
```
marcoscaceres: But what happens if the you do this form an iframe, navigate the iframe away, which then settles the promise (and presumedly also aborts the download)?

Also, it seems unnecessary to not allow aborting the download if you make (or fake) all downloads so they are origin bound, even if they are a shared component, right? ... that is, if all sites MUST attempt to download a resource before they can use it.

marcoscaceres: Ah, iframes are dealt with in the next paragraph... I see..

index.bs
```diff
@@ -136,6 +139,116 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+Once a download is initiated by {{SpeechRecognition/install()}}, the user agent should preserve the download progress. <strong>User agents should not cancel an ongoing language pack download in response to page-controlled actions</strong> (e.g., navigation, page unload) that could be used to manipulate the download state for fingerprinting. If a page navigates away, the download should ideally continue in the background, or at least its progress should be saved. The goal is to prevent the site from easily reverting a language pack's state from {{"downloading"}} back to {{"downloadable"}}.
```
marcoscaceres: This honestly seems like not a great way to mitigate a larger problem... at least, this doesn't feel right.

index.bs
evanbliu: Yes, that's how it is implemented in Chromium and how this spec is supposed to be written. I've made some changes here clarifying that.

index.bs
evanbliu: I've updated this to mandate the same masking scheme already used by Chromium.

index.bs
evanbliu: Since availability is now strictly partitioned by storage key to prevent cross-site fingerprinting, forcing background downloads is no longer necessary. The spec has been updated to allow browsers to cancel downloads normally when a page closes.

index.bs
```diff
@@ -136,6 +139,118 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+<h5 id="on-device-model-privacy-availability-eviction">Download Eviction</h5>
+
+Another ingredient in ensuring that websites cannot toggle the availability state back and forth is to ensure that user agents don't use a quota-based eviction system for downloaded language packs that web pages can indirectly control. For example, if a user agent evicted less-recently-used language packs when new ones are installed, a web page could trigger such evictions to toggle the state of a target language pack.
+
+To avoid this, <strong>user agents should not implement systems which allow web pages to control the eviction of downloaded language packs</strong>, including via indirect triggers such as further subsequent downloads. One way to fulfill this requirement is to never evict downloaded material in response to web page-initiated storage pressure, instead refusing to download new material (e.g., {{SpeechRecognition/install()}} resolving to `false`) if doing so would cause storage pressure.
+
+Evicting downloads in response to user-controlled actions (e.g., via a browser settings UI) is not problematic.
+
+<h5 id="on-device-model-privacy-availability-alternatives">Alternate Options</h5>
```
hoch: Perhaps I am missing something, but:

In the "Download Masking" section above, partitioning availability by `[=storage key=]` was made a normative **must** requirement in the latest commit. However, this section still frames storage key partitioning as an alternative option with trade-offs.

Should we update or remove this subsection so it doesn't contradict the mandatory partitioning requirement in the masking section?

index.bs
```diff
@@ -136,6 +139,118 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+Implementations <strong>must</strong> ensure that this [=storage key=]-based masking state is cleared whenever other storage for that origin is reset by the user.
+
+<h5 id="on-device-model-privacy-availability-installation">Installation-time friction</h5>
+
+The mitigation described in <a href="#on-device-model-privacy-availability-masking">Download Masking</a> works against attempts to silently fingerprint using {{SpeechRecognition/available()}}. The specification also contains requirements to prevent {{SpeechRecognition/install()}} from being easily used for fingerprinting, by introducing friction:
+
+*   The {{SpeechRecognition/install()}} method both requires and consumes [=user activation=], when it would initiate a download.
+*   The {{SpeechRecognition/install()}} method allows the user agent to prompt the user for permission, or to implicitly reject download attempts based on previous signals (such as an observed pattern of abuse).
+*   Access to {{SpeechRecognition/install()}} and {{SpeechRecognition/available()}} is gated on an per-API [=policy-controlled feature=], which means that only top-level origins and their delegates can use the API.
```
hoch: typo: `an per-API` -> `a per-API`

index.bs
```diff
@@ -136,6 +139,118 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+<h4 id="on-device-model-privacy-availability-masking">Download Masking</h4>
+
+To prevent cross-origin fingerprinting and site collusion, user agents <strong>must</strong> partition language pack availability by [=storage key=]. User agents <strong>must</strong> mask the current download status by returning {{"downloadable"}} from {{SpeechRecognition/available()}} by default, even if the actual underlying language pack is {{"available"}} or {{"downloading"}} on the user's device.
+
+This masking state is established for each (API, options, [=storage key=]) tuple. The user agent <strong>must only</strong> set the state to "unmasked" for a specific [=storage key=] after a web page within that key explicitly calls {{SpeechRecognition/install()}} with a given set of options, and successfully starts a download or the promise resolves to <code>true</code> (indicating the language pack is ready).
```
hoch: formatting nit: `<code>true</code>` -> `` `true` ``

hoch (commented):
Mostly nits. Looking good!

index.bs
```diff
@@ -136,6 +139,108 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+Since {{SpeechRecognition/install()}} has stronger requirements (see <a href="#on-device-model-privacy-availability-installation">Installation-time friction</a>), this ensures that web pages only get access to the true download status after taking a more costly action, such as triggering a user permission prompt.
+
+Implementations <strong>must</strong> ensure that this [=storage key=]-based masking state is cleared whenever other storage for that origin is reset by the user.
+
+<h5 id="on-device-model-privacy-availability-installation">Installation-time friction</h5>
+
+The mitigation described in <a href="#on-device-model-privacy-availability-masking">Download Masking</a> works against attempts to silently fingerprint using {{SpeechRecognition/available()}}. The specification also contains requirements to prevent {{SpeechRecognition/install()}} from being easily used for fingerprinting, by introducing friction:
+
+*   The {{SpeechRecognition/install()}} method both requires and consumes [=user activation=], when it would initiate a download.
```
hoch, line 163: I noticed we note here that `install()` requires and consumes user activation, but looking at the algorithm for `SpeechRecognition.install()` below, it doesn't seem to have steps checking or consuming transient activation yet.

What do you think about aligning the two? Would you prefer adding those steps to `install()` in this PR, or should we track that in a quick follow-up issue?

index.bs
```diff
@@ -136,6 +139,108 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+<h4 id="on-device-model-privacy-availability-masking">Download Masking</h4>
+
+To prevent cross-origin fingerprinting and site collusion, user agents <strong>must</strong> partition language pack availability by [=storage key=]. User agents <strong>must</strong> mask the current download status by returning {{"downloadable"}} from {{SpeechRecognition/available()}} by default, even if the actual underlying language pack is {{"available"}} or {{"downloading"}} on the user's device.
+
+This masking state is established for each (API, options, [=storage key=]) tuple. The user agent <strong>must only</strong> set the state to "unmasked" for a specific [=storage key=] after a web page within that key explicitly calls {{SpeechRecognition/install()}} with a given set of options, and successfully starts a download or the promise resolves to `true` (indicating the language pack is ready).
```
hoch, line 153: Since `SpeechRecognition` is the only interface managing model downloads here, do you think we still need `API` in this tuple? Wondering if `(options, [=storage key=])` might be slightly more direct for readers.

index.bs
```diff
@@ -136,6 +139,108 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+Implementations <strong>must</strong> ensure that this [=storage key=]-based masking state is cleared whenever other storage for that origin is reset by the user.
+
+<h5 id="on-device-model-privacy-availability-installation">Installation-time friction</h5>
+
+The mitigation described in <a href="#on-device-model-privacy-availability-masking">Download Masking</a> works against attempts to silently fingerprint using {{SpeechRecognition/available()}}. The specification also contains requirements to prevent {{SpeechRecognition/install()}} from being easily used for fingerprinting, by introducing friction:
+
+*   The {{SpeechRecognition/install()}} method both requires and consumes [=user activation=], when it would initiate a download.
+*   The {{SpeechRecognition/install()}} method allows the user agent to prompt the user for permission, or to implicitly reject download attempts based on previous signals (such as an observed pattern of abuse).
+*   Access to {{SpeechRecognition/install()}} and {{SpeechRecognition/available()}} is gated on a per-API [=policy-controlled feature=], which means that only top-level origins and their delegates can use the API.
```
hoch, line 165: Since both methods are specifically gated behind "on-device-speech-recognition", would it be clearer to name the feature directly here (e.g. "on-device-speech-recognition") rather than referring to a "per-API" feature?

index.bs
```diff
@@ -136,6 +139,108 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+<strong>If a previously installed language pack is evicted (e.g., by the user or due to extreme storage pressure) while it might be in use or expected to be available, subsequent attempts to use it (e.g., via {{SpeechRecognition/start()}} with {{SpeechRecognition/lang}} set to that language and {{SpeechRecognition/processLocally}} as true) should fail gracefully. This might involve {{SpeechRecognition/available()}} returning {{"downloadable"}} or {{"unavailable"}}, and {{SpeechRecognition/start()}} potentially firing an {{SpeechRecognitionErrorEvent}} with an appropriate error code like {{SpeechRecognitionErrorCode/language-not-supported}} or {{SpeechRecognitionErrorCode/service-not-allowed}}.</strong>
```
hoch, line 228: A couple of small phrasing nits here:

- Would `set to true` be clearer than `as true` for the boolean attribute?
- Should this be `a {{SpeechRecognitionErrorEvent}}` rather than `an`?

index.bs
```diff
@@ -136,6 +139,108 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+<h4 id="on-device-model-security-runtime-resources">Runtime Shared Resources</h4>
+
+On-device speech recognition can consume significant runtime resources like CPU, memory, and potentially specialized hardware accelerators.
+
+<strong>User agents should ensure that one web page's use of on-device speech recognition does not overly interfere with another web page's use of the API, or another web page's general operation, or the overall system stability.</strong> For example, it should not be possible for a background tab to monopolize speech processing resources, preventing a foreground tab from using them.
+
+This specification does not mandate any particular mitigation strategy, but possible approaches include queuing requests, rate limiting, prioritizing foreground tabs, or detecting abusive behavior. <strong>If necessary to prevent resource exhaustion or instability, the user agent may cause speech recognition operations to fail (e.g., by firing an {{SpeechRecognitionErrorEvent}} with {{SpeechRecognitionErrorCode/service-not-allowed}}).</strong>
```
hoch, line 236: Should this be `a {{SpeechRecognitionErrorEvent}}`?

index.bs
```diff
@@ -136,6 +139,108 @@ This does not preclude adding support for this as a future API enhancement, and
...
+
+<h4 id="on-device-model-security-os-models">OS-Provided Models</h4>
+
+One implementation strategy for on-device speech recognition is to delegate to models or capabilities provided by the underlying operating system. This can offer benefits like a consistent user experience and efficient resource usage.
+
+However, this approach comes with the usual considerations of exposing OS capabilities to the web. User agents must still ensure that all privacy and security requirements of this specification are met when using OS-provided models. This includes the requirements in <a href="#on-device-model-privacy-user-input">User Input and Speech Data</a> (preventing training on user data) and <a href="#on-device-model-security-runtime-resources">Runtime Shared Resources</a> (ensuring fair and stable resource sharing).
```
hoch, line 242: Is this section non-normative? Otherwise I think we should use standard RFC 2119 uppercase `MUST` here.

hoch (commented):
Mostly nits, but I do have some questions.

PR #169, opened 2025-06-23
Replace SpeechRecognitionPhraseList with a sequence

### LINKED ISSUE #168: Consider replacing SpeechRecognitionPhraseList with sequence<SpeechRecognitionPhrase>
SpeechRecognitionPhraseList was added to mirror pattern established by SpeechGrammarList, but it would be simpler to use a sequence<SpeechRecognitionPhrase> instead of creating a new type.

evanbliu:
This PR replaces SpeechRecognitionPhraseList with sequence<SpeechRecognitionPhrase>.

Closes #168

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -163,7 +163,7 @@ The term "interim result" indicates a {{SpeechRecognitionResult}} in which the {
 <dl dfn-type=attribute dfn-for="SpeechRecognition">
     : <dfn>[[phrases]]</dfn>
     ::
-        A {{SpeechRecognitionPhraseList}} representing a list of phrases for contextual biasing. The initial value is null.
+        A sequence of {{SpeechRecognitionPhrase}} representing a list of phrases for contextual biasing. The initial value is an empty sequence.
 </dl>
 
 <xmp class="idl">
@@ -178,7 +178,7 @@ interface SpeechRecognition : EventTarget {
     attribute boolean interimResults;
     attribute unsigned long maxAlternatives;
     attribute boolean processLocally;
-    attribute SpeechRecognitionPhraseList phrases;
+    attribute sequence<SpeechRecognitionPhrase> phrases;
 
     // methods to drive the speech interaction
     undefined start();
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -178,7 +178,7 @@ interface SpeechRecognition : EventTarget {
     attribute boolean interimResults;
     attribute unsigned long maxAlternatives;
     attribute boolean processLocally;
-    attribute SpeechRecognitionPhraseList phrases;
+    attribute sequence<SpeechRecognitionPhrase> phrases;
```
domenic, line 181: This is invalid Web IDL; attributes cannot be sequences.

ObservableArray is probably the correct choice here. See https://webidl.spec.whatwg.org/#idl-observable-array

evanbliu:
On second thought, I'm questioning if this is the right thing to do here for the following reasons:
1. The browser won't be modifying the SpeechRecognitionPhrases, so there's no need for websites to observe changes to this collection and using ObservableArray would imply that the browser's speech engine might add or remove the phrases in the list.
2. I didn't find any other examples of APIs with ObservableArrays as attributes. In Chromium, there's only [one](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/document_or_shadow_root.idl?q=ObservableArray%20-f:out%20f:%5C.idl%20-f:test&ss=chromium%2Fchromium%2Fsrc) non-test usage of ObservableArray.
3. There's already a precedent of creating a new interface type with the sole purpose of encapsulating a collection (e.g. [SpeechRecognitionPhraseList](https://webaudio.github.io/web-speech-api/#speechreco-phraselist), [TouchList](https://w3c.github.io/touch-events/#touchlist-interface), [FileList](https://w3c.github.io/FileAPI/#dfn-filelist)).

My suggestion is to keep the SpeechRecognitionPhraseList as-is, or alternatively use a [FrozenArray](https://webidl.spec.whatwg.org/#idl-frozen-array) it's crucial to avoid adding a new type. @domenic - What are your thoughts on this?

domenic:
It's indeed crucial to avoid adding new fake arrays to the web platform. That is a strongly deprecated practice.

The question is whether you want this to be mutable, or only settable by the constructor. If it's supposed to be mutable, ObservableArray is the right choice. If it's only settable by the constructor, then adding a constructor option (that would be a `sequence<>`) and changing the type to `readonly FrozenArray<T>` makes sense.

Some recent discussion about this in https://github.com/whatwg/webidl/issues/1495#issuecomment-2975168470.

evanbliu:
I see, it sounds like ObservableArray would be the best choice here since it's mutable.

Do you know why there's still only a [single](https://source.chromium.org/search?q=%22observable_array%22%20f:.gni&ss=chromium%2Fchromium%2Fsrc) usage of the ObservableArray in Chromium despite being introduced 4 years ago?

domenic:
I think it's a matter of successive filtering. You'd need:

- To produce a new API
- That needs an array-like type
- That fits into cases (2) or (3) of my linked comment, instead of (1) or (4).

I think it's pretty reasonable to expect only one API to meet those filters over the course of 4 years.

PR #192, opened 2026-07-14
Update SpeechRecognitionResult with speech timing attributes

### LINKED ISSUE #191: Add SpeechRecognitionResult Timestamps
### Description

The Web Speech API currently does not expose the start and end timestamps of the source audio corresponding to a transcription result (`SpeechRecognitionResult`). 

While `SpeechRecognitionEvent.timeStamp` provides a single timestamp for the event itself, it does not convey the **start time, end time, or duration of the acoustic speech segment within the audio stream**. 

Without acoustic bounds for each result, web applications cannot associate transcribed text with media timelines:

- **Automated Subtitling & Caption Tracks:** Cannot construct frame-accurate subtitle cues (`VTTCue(start, end, text)`) for recorded video or audio.
- **Interactive "Click-to-Seek" Transcripts:** Cannot jump media playback (`mediaElement.currentTime = result.speechStartTime`) to the exact moment a phrase was spoken.
- **Live Video Conferencing & Recording Sync:** In applications like Google Meet or WebRTC calls, captions displayed live cannot be accurately aligned with the recorded video track for post-meeting review or synchronized with video playout delay buffers.
- **Text-Based Media Editing:** Tools that allow users to edit audio/video by cutting or rearranging transcript text require precise acoustic boundaries for each segment.

---

### Specification

Expose speech segment bounds on the [`SpeechRecognitionResult`](https://webaudio.github.io/web-speech-api/#speechreco-result) interface in **seconds** as a `double`, measured on the timeline of the audio stream consumed by the recognizer ($t = 0.0\text{s}$, marked by `audiostart`).

#### Web IDL

```webidl
partial interface SpeechRecognitionResult {
    // Start timestamp of the recognized speech segment in seconds relative to the audio stream (0.0s).
    readonly attribute double speechStartTime;

    // End timestamp of the recognized speech segment in seconds relative to the audio stream.
    readonly attribute double speechEndTime;
};
```

---

### Usage Example: Subtitling & "Click-to-Seek" Navigation

```javascript
const mediaElement = document.querySelector('video');
const track = mediaElement.addTextTrack('captions', 'English', 'en');
track.mode = 'showing';

const recognition = new SpeechRecognition();
recognition.continuous = true;
recognition.interimResults = false;

recognition.onresult = (event) => {
  for (let i = event.resultIndex; i < event.results.length; ++i) {
    const result = event.results[i];
    if (!result.isFinal) continue;

    const transcriptText = result[0].transcript;

    // 1. Create frame-accurate WebVTT cue directly from seconds
    const cue = new VTTCue(result.speechStartTime, result.speechEndTime, transcriptText);
    track.addCue(cue);

    // 2. Click-to-seek transcript element
    const item = document.createElement('p');
    item.textContent = `[${result.speechStartTime.toFixed(1)}s] ${transcriptText}`;
    item.onclick = () => {
      mediaElement.currentTime = result.speechStartTime;
      mediaElement.play();
    };
    document.getElementById('transcript').appendChild(item);
  }
};

recognition.start(mediaElement.captureStream().getAudioTracks()[0]);
```

---

### Security & Privacy Considerations

- **Fingerprinting Mitigation:** To mitigate micro-architectural timing attacks and hardware fingerprinting, user agents may reduce the resolution of `speechStartTime` and `speechEndTime` or introduce jitter, in accordance with user agent security and privacy policies (consistent with [[HR-TIME-3]] and [[HTML]]).

---

### Alternatives Considered

- **Existing `SpeechRecognitionEvent.timeStamp`:** Represents a single point in time (the event occurrence in document time). It does not provide the duration or start/end acoustic boundaries of the utterance within the stream, making subtitle cue generation (`VTTCue`) and click-to-seek impossible.
- **Separate Events (`speechstart` / `speechend`):** Fire only once at session voice-activity boundaries, cannot be correlated to individual stream results in continuous recognition, and do not provide segment start/end bounds per result.
- **`DOMHighResTimeStamp` (Milliseconds):** Semantically represents time relative to document origin rather than a media stream timeline, and causes friction with existing W3C media APIs (`HTMLMediaElement.currentTime`, `BaseAudioContext.currentTime`, and `VTTCue`) which universally operate in seconds as a `double`.

---

### Links & References

- **Explainer:** [speech-recognition-result-timestamps.md](https://github.com/WebAudio/web-speech-api/blob/main/explainers/speech-recognition-result-timestamps.md) ([PR #206](https://github.com/WebAudio/web-speech-api/pull/206))
- **Spec PR:** [PR #192](https://github.com/WebAudio/web-speech-api/pull/192)
- **Interactive Demos:**
  - [Live Meeting Caption Sync & Recording Demo](https://alan33d.github.io/web-speech-demos/live_meeting_recording_demo.html)
  - [WebRTC Caption Playout Delay Test](https://alan33d.github.io/web-speech-demos/webrtc_caption_sync_test.html)
- **ChromeStatus:** [Web Speech API: Add SpeechRecognitionResult Timestamps](https://chromestatus.com/feature/5811907077472256)

alan33d:
Added speechStartTime and speechEndTime attributes to SpeechRecognitionResult interface with explanations as proposed in #191.

Fixes: #191

***
<a href="https://pr-preview.s3.amazonaws.com/alan33d/web-speech-api/pull/192.html" title="Last updated on Sep 11, 2026, 5:32 AM UTC (e3cf78c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/192/291cb48...alan33d:e3cf78c.html" title="Last updated on Sep 11, 2026, 5:32 AM UTC (e3cf78c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -364,10 +364,10 @@ interface SpeechRecognitionPhrase {
 
 <dl>
   <dt><dfn attribute for=SpeechRecognitionResult>audioStartTime</dfn> attribute</dt>
-  <dd>A nullable {{DOMHighResTimeStamp}} representing the start of the audio segment corresponding to this recognition result, in milliseconds relative to the time origin. Returns null if the underlying recognition engine does not support audio segment start timestamps.</dd>
+  <dd>A nullable {{DOMHighResTimeStamp}} representing the start of the audio segment corresponding to this recognition result, in milliseconds relative to the start of the audio stream. Returns null if the underlying recognition engine does not support audio segment start timestamps.</dd>
 
   <dt><dfn attribute for=SpeechRecognitionResult>audioEndTime</dfn> attribute</dt>
-  <dd>A nullable {{DOMHighResTimeStamp}} representing the end of the audio segment corresponding to this recognition result, in milliseconds relative to the time origin. Returns null if the underlying recognition engine does not support audio segment end timestamps.</dd>
+  <dd>A nullable {{DOMHighResTimeStamp}} representing the end of the audio segment corresponding to this recognition result, in milliseconds relative to the start of the audio stream. Returns null if the underlying recognition engine does not support audio segment end timestamps.</dd>
 </dl>
 
 <p class=issue>The group has discussed whether WebRTC might be used to specify selection of audio sources and remote recognizers.
--- a/index.bs
+++ b/index.bs
@@ -121,7 +121,7 @@ This does not preclude adding support for this as a future API enhancement, and 
 
   <li>To mitigate the risk of fingerprinting, user agents MUST NOT personalize speech recognition when performing speech recognition on a {{MediaStreamTrack}}.</li>
 
-  <li>To mitigate fingerprinting vectors associated with high-precision timing, user agents MUST apply timestamp fuzzing and precision reduction to {{SpeechRecognitionResult/audioStartTime}} and {{SpeechRecognitionResult/audioEndTime}} before exposing these attributes to scripts (e.g. by rounding to 2ms precision).</li>
+  <li>To mitigate micro-architectural timing attacks and hardware fingerprinting, user agents may reduce the resolution of {{SpeechRecognitionResult/audioStartTime}} and {{SpeechRecognitionResult/audioEndTime}} or introduce jitter, in accordance with the user agent's security and privacy policies (similar to [[HR-TIME-3]] and [[HTML]]).</li>
 </ol>
 
 <h3 id="implementation-considerations">Implementation considerations</h3>
@@ -260,8 +260,8 @@ interface SpeechRecognitionResult {
     readonly attribute unsigned long length;
     getter SpeechRecognitionAlternative? item(unsigned long index);
     readonly attribute boolean isFinal;
-    readonly attribute DOMHighResTimeStamp? audioStartTime;
-    readonly attribute DOMHighResTimeStamp? audioEndTime;
+    readonly attribute double audioStartTime;
+    readonly attribute double audioEndTime;
 };
 
 // A collection of responses (used in continuous mode)
@@ -364,10 +364,10 @@ interface SpeechRecognitionPhrase {
 
 <dl>
   <dt><dfn attribute for=SpeechRecognitionResult>audioStartTime</dfn> attribute</dt>
-  <dd>A nullable {{DOMHighResTimeStamp}} representing the start of the audio segment corresponding to this recognition result, in milliseconds relative to the start of the audio stream. Returns null if the underlying recognition engine does not support audio segment start timestamps.</dd>
+  <dd>A {{double}} representing the start time of the audio segment corresponding to this recognition result, in seconds relative to the start of the audio stream consumed by the speech recognizer.</dd>
 
   <dt><dfn attribute for=SpeechRecognitionResult>audioEndTime</dfn> attribute</dt>
-  <dd>A nullable {{DOMHighResTimeStamp}} representing the end of the audio segment corresponding to this recognition result, in milliseconds relative to the start of the audio stream. Returns null if the underlying recognition engine does not support audio segment end timestamps.</dd>
+  <dd>A {{double}} representing the end time of the audio segment corresponding to this recognition result, in seconds relative to the start of the audio stream consumed by the speech recognizer.</dd>
 </dl>
 
 <p class=issue>The group has discussed whether WebRTC might be used to specify selection of audio sources and remote recognizers.
--- a/index.bs
+++ b/index.bs
@@ -121,7 +121,7 @@ This does not preclude adding support for this as a future API enhancement, and 
 
   <li>To mitigate the risk of fingerprinting, user agents MUST NOT personalize speech recognition when performing speech recognition on a {{MediaStreamTrack}}.</li>
 
-  <li>To mitigate micro-architectural timing attacks and hardware fingerprinting, user agents may reduce the resolution of {{SpeechRecognitionResult/speechStartTime}} and {{SpeechRecognitionResult/speechEndTime}} or introduce jitter, in accordance with the user agent's security and privacy policies (similar to [[HR-TIME-3]] and [[HTML]]).</li>
+  <li>To mitigate micro-architectural timing attacks and hardware fingerprinting, user agents MAY reduce the resolution of {{SpeechRecognitionResult/speechStartTime}} and {{SpeechRecognitionResult/speechEndTime}} or introduce jitter, in accordance with the user agent's security and privacy policies (similar to [[HR-TIME-3]] and [[HTML]]).</li>
 </ol>
 
 <h3 id="implementation-considerations">Implementation considerations</h3>
@@ -358,16 +358,6 @@ interface SpeechRecognitionPhrase {
   <dd>
     The getter steps are to return the value of {{SpeechRecognition/[[phrases]]}}.
   </dd>
-</dl>
-
-<h4 id="speechrecoresult-attributes">SpeechRecognitionResult Attributes</h4>
-
-<dl>
-  <dt><dfn attribute for=SpeechRecognitionResult>speechStartTime</dfn> attribute</dt>
-  <dd>A {{double}} representing the start time of the speech segment corresponding to this recognition result, in seconds relative to the start of the audio stream consumed by the speech recognizer (where 0.0 seconds corresponds to the {{SpeechRecognition/onaudiostart}} event).</dd>
-
-  <dt><dfn attribute for=SpeechRecognitionResult>speechEndTime</dfn> attribute</dt>
-  <dd>A {{double}} representing the end time of the speech segment corresponding to this recognition result, in seconds relative to the start of the audio stream consumed by the speech recognizer.</dd>
 </dl>
 
 <p class=issue>The group has discussed whether WebRTC might be used to specify selection of audio sources and remote recognizers.
@@ -680,6 +670,12 @@ For example, some implementations may fire <a event for=SpeechRecognition>audioe
   <dt><dfn attribute for=SpeechRecognitionResult>isFinal</dfn> attribute</dt>
   <dd>The final boolean must be set to true if this is the final time the speech service will return this particular index value.
   If the value is false, then this represents an interim result that could still be changed.</dd>
+
+  <dt><dfn attribute for=SpeechRecognitionResult>speechStartTime</dfn> attribute</dt>
+  <dd>A {{double}} representing the start time of the speech segment corresponding to this recognition result, in seconds relative to the start of the audio stream consumed by the speech recognizer (where 0.0 seconds corresponds to the <a event for=SpeechRecognition>audiostart</a> event).</dd>
+
+  <dt><dfn attribute for=SpeechRecognitionResult>speechEndTime</dfn> attribute</dt>
+  <dd>A {{double}} representing the end time of the speech segment corresponding to this recognition result, in seconds relative to the start of the audio stream consumed by the speech recognizer.</dd>
 </dl>
 
 <h4 id="speechreco-resultlist">SpeechRecognitionResultList</h4>
--- a/index.bs
+++ b/index.bs
@@ -121,7 +121,7 @@ This does not preclude adding support for this as a future API enhancement, and 
 
   <li>To mitigate the risk of fingerprinting, user agents MUST NOT personalize speech recognition when performing speech recognition on a {{MediaStreamTrack}}.</li>
 
-  <li>To mitigate micro-architectural timing attacks and hardware fingerprinting, user agents may reduce the resolution of {{SpeechRecognitionResult/audioStartTime}} and {{SpeechRecognitionResult/audioEndTime}} or introduce jitter, in accordance with the user agent's security and privacy policies (similar to [[HR-TIME-3]] and [[HTML]]).</li>
+  <li>To mitigate micro-architectural timing attacks and hardware fingerprinting, user agents may reduce the resolution of {{SpeechRecognitionResult/speechStartTime}} and {{SpeechRecognitionResult/speechEndTime}} or introduce jitter, in accordance with the user agent's security and privacy policies (similar to [[HR-TIME-3]] and [[HTML]]).</li>
 </ol>
 
 <h3 id="implementation-considerations">Implementation considerations</h3>
@@ -260,8 +260,8 @@ interface SpeechRecognitionResult {
     readonly attribute unsigned long length;
     getter SpeechRecognitionAlternative? item(unsigned long index);
     readonly attribute boolean isFinal;
-    readonly attribute double audioStartTime;
-    readonly attribute double audioEndTime;
+    readonly attribute double speechStartTime;
+    readonly attribute double speechEndTime;
 };
 
 // A collection of responses (used in continuous mode)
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -258,6 +260,8 @@ interface SpeechRecognitionResult {
     readonly attribute unsigned long length;
     getter SpeechRecognitionAlternative item(unsigned long index);
     readonly attribute boolean isFinal;
+    readonly attribute DOMHighResTimeStamp? audioStartTime;
```
hoch: I am curious what these timestamps mean when `isFinal` is `false`.

For example, when an interim result is updated with more speech (e.g., 'hello' becomes 'hello world'), does audioEndTime update dynamically? Does audioStartTime remain stable for the duration of that interim result's lifecycle?

index.bs
```diff
@@ -356,6 +360,16 @@ interface SpeechRecognitionPhrase {
   </dd>
 </dl>
 
+<h4 id="speechrecoresult-attributes">SpeechRecognitionResult Attributes</h4>
+
+<dl>
+  <dt><dfn attribute for=SpeechRecognitionResult>audioStartTime</dfn> attribute</dt>
+  <dd>A nullable {{DOMHighResTimeStamp}} representing the start of the audio segment corresponding to this recognition result, in milliseconds relative to the time origin. Returns null if the underlying recognition engine does not support audio segment start timestamps.</dd>
```
hoch: IIRC `SpeechSynthesisEvent` uses `elapsedTime`, which is `float`.

Is there a specific design reason for using DOMHighResTimeStamp here?

alan33d: No specific reason, I started with `double?` but it was suggested by @evanbliu to use DOMHighResTimeStamp? to mirror `event.timeStamp` property.   I don't have a strong preference on this.

evanbliu: IIUC, `DOMHighResTimeStamp` is just a typedef for a `double` but it makes it clear that the units are in milliseconds.

I assume `SpeechSynthesisEvent` uses `float` because `DOMHighResTimeStamp` didn't exist (or at least wasn't widely adopted) back in 2012.

index.bs
alan33d: >  does audioEndTime update dynamically?

Yes, for a given SpeechRecognitionResult lifecycle (while isFinal is false), audioEndTime advances to reflect the end boundary of the audio buffer processed by the ASR engine so far.   

When the ASR engine detects an endpoint and sets isFinal to true, subsequent speech will create a new result index with a new audioStartTime marking the onset of the next speech segment.

> Does audioStartTime remain stable for the duration of that interim result's lifecycle?

- audioStartTime remains stable throughout the interim updates, anchoring the start timestamp of the utterance segment.
- audioEndTime updates dynamically with each interim event as additional audio is processed by the engine.
- Once isFinal becomes true, both timestamps freeze for that result item.

I also double checked SODA's behavior of `audio_start_time` in [1] which I plan on plumbing into this param, including `TimingMetrics` proto (speech/soda/public/soda_event.proto) and SODA session handling (soda_sync_session.cc). Throughout a series of `PartialRecognitionResult ` events (isFinal == false), SODA does maintain the same audio_start_time baseline for that un-finalized segment.

[1]https://source.chromium.org/chromium/chromium/src/+/main:media/mojo/mojom/speech_recognition_result.h;drc=0ca5fe6d5cc9a68ac98b6f5d82174b5d40009a10;bpv=1;bpt=1;l=62

evanbliu:
No objections from me, but I’ll defer the review part to @padenot to make sure we have browser alignment here.

index.bs
```diff
@@ -120,6 +120,8 @@ This does not preclude adding support for this as a future API enhancement, and
   <li>The user agent may also give the user a longer explanation the first time speech input is used, to let the user know what it is and how they can tune their privacy settings to disable speech recording if required.</li>
 
   <li>To mitigate the risk of fingerprinting, user agents MUST NOT personalize speech recognition when performing speech recognition on a {{MediaStreamTrack}}.</li>
+
+  <li>To mitigate fingerprinting vectors associated with high-precision timing, user agents MUST apply timestamp fuzzing and precision reduction to {{SpeechRecognitionResult/audioStartTime}} and {{SpeechRecognitionResult/audioEndTime}} before exposing these attributes to scripts (e.g. by rounding to 2ms precision).</li>
```
padenot: This isn't a useful mitigation is it? Think about how you'd implement the same ASR system in WASM, will it be limited? It's absolutely _trivial_ to implement `audioStartTime` etc. in WASM, and it will be sample-accurate w.r.t the audio stream.

index.bs
```diff
@@ -356,6 +360,16 @@ interface SpeechRecognitionPhrase {
   </dd>
 </dl>
 
+<h4 id="speechrecoresult-attributes">SpeechRecognitionResult Attributes</h4>
+
+<dl>
+  <dt><dfn attribute for=SpeechRecognitionResult>audioStartTime</dfn> attribute</dt>
+  <dd>A nullable {{DOMHighResTimeStamp}} representing the start of the audio segment corresponding to this recognition result, in milliseconds relative to the start of the audio stream. Returns null if the underlying recognition engine does not support audio segment start timestamps.</dd>
```
padenot: We use seconds throughout those APIs, or microseconds in codec APIs. 64-bits floating point numbers have enough resolution here at even high sample rates, and it means we're not going to convert to/from seconds.

I don't think those should be nullalble. If your backend doesn't support timestamping, it can be implemented, or you can ship something decent (in the sense that it's going to be accurate enough for a lot of use case) in front of it.

We tend (not 100%)  to not specificy APIs because of certain technologies limitations. Instead, we improve those technologies and write a good spec.

As discussed in the other PR, "relative to the start of the audio stream" is very fuzzy. How does it work if you get a `MediaStream` with a certain `currentTime` (internally), then fed into the Web Audio API (which has a different `currentTime`), then fed into an `HTMLMediaElement` (different `currentTime`), then `captureStream()`, and piped into `SpeechRecognition`. Which it is? The last one? Probably, and in that clock domain as well, which is important to mention because crossing clock domain makes alignment hard (which is what we want to do here).

Why is this not `speechStartTime`, mirroring `speechStart` (that also has a timestamp). What are the constraints between the two? It should be strictly equal probably?

padenot (commented):
Some questions below, thanks.

index.bs
alan33d: I see your point regarding WASM and updated this section. IIUC, any web application with microphone access can already feed raw audio frames to an on-device ASR model compiled to WASM/AudioWorklet and obtain sample-accurate boundary timestamps.

We agree that **the W3C specification itself should not mandate a specific fuzzing constraint in the text**. 
Following the precedent set by the [W3C High Resolution Time Level 3](https://www.w3.org/TR/hr-time-3/#privacy-security) and HTML specifications, the spec can define `audioStartTime` and `audioEndTime` as accurate representations of the audio stream timing, while leaving timestamp fuzzing to the discretion of each browser's security requirements.

Just for context, the 2ms discrete quantization was introduced to align with the existing security posture established by [`HTMLMediaElement.currentTime`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime) and `DOMHighResTimeStamp`. 
Exposing sub-millisecond, fine-grained hardware timing in web platform APIs has historically raised security review concerns regarding side-channel attacks and hardware fingerprinting.

index.bs
alan33d: Thanks @padenot for the insights!

**1. Units (Seconds vs. Milliseconds):**
We currently return milliseconds (as a `double`), matching `DOMHighResTimeStamp`, `performance.now()`, and `event.timeStamp` (https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp). We're happy to align with the WG's preference if seconds (like Web Audio `AudioContext.currentTime`) is preferred.  See updates.

**2. Non-nullable Attributes:**
Fair point on specifying a clean API rather than designing around backend limitations. We've updated the spec to make these non-nullable.

Rather than returning sentinel values like `0` or `-1` when a backend does not support timestamps, we leave this case out of the specification entirely. If an implementation or backend doesn't support timestamps, the attribute simply should not be present. We can work through the Chrome-internal details of ensuring the attributes are only exposed when supported by the underlying Web Speech backend.

**3. `audioStartTime` vs. `speechstart` / Why they are not strictly equal:**
* In continuous recognition, `speechstart` fires once when voice activity is first detected in the session. A continuous session lasting several minutes will produce dozens of `SpeechRecognitionResult` items over time. For example, a result at minute 2 has `audioStartTime = 120000ms`, whereas `speechstart` fired at $t = 1200ms$. They are only equal for the very first utterance.
* `audioStartTime` and `audioEndTime` represent the acoustic sample window in the audio stream associated with that specific transcript hypothesis.
* Correlating separate `speechstart` / `speechend` DOM events with streaming interim and final `SpeechRecognitionResult` objects creates a fragile state-machine burden for developers. Attaching timestamps directly to `SpeechRecognitionResult` provides an atomic, ergonomic association.
* Furthermore, having `result.audioEndTime` alongside `event.timeStamp` (which reflects when the DOM event fired) allows developers to compute engine processing latency: `event.timeStamp - result.audioEndTime`.

We have also updated the Explainer to detail point 3: https://github.com/WebAudio/web-speech-api/pull/205.

padenot: The standard is quite clear that `event.timeStamp` isn't when the event fired:

> For all these events, the timeStamp attribute defined in the DOM Level 2 Event interface must be set to the best possible estimate of when the real-world event which the event object represents occurred. This timestamp must be represented in the user agent’s view of time, even for events where the timestamps in question could be raised on a different machine like a remote recognition service (i.e., in a [speechend](https://webaudio.github.io/web-speech-api/#eventdef-speechrecognition-speechend) event with a remote speech endpointer).

padenot: for `2.`, the a compliant implementation cannot leave those attributes out. When an implementation doesn't support timestamp, it needs to be improved to support them to be compliant.

alan33d: You are right on `event.timeStamp`, and Chromium had a previously existing Blink bug where the timestamp is set to `base::TimeTicks::Now()` when the event is dispatched rather than the end of the recognition.  We are actively fixing this and agree it can be used for latency tracking.

We have updated the PR and Spec issue to incorporate your feedback:
- Renamed the attributes from `audio*` to `speechStartTime` and `speechEndTime` to mirror `speechstart` / `speechend` and clarify that these timestamps delimit the recognized speech utterance.
- We also updated the explainer in #206 to focus on timeline association use cases, associating transcripts which uses start and end recognition timestamps (generating `VTTCue` subtitle tracks, "click-to-seek" transcript navigation, recorded meeting alignment, and text-based media editing).

alan33d:
PTALA, thanks!

reillyeon (approved):

index.bs
```diff
@@ -258,6 +260,8 @@ interface SpeechRecognitionResult {
     readonly attribute unsigned long length;
     getter SpeechRecognitionAlternative item(unsigned long index);
```
hoch: Perhaps we need to rebase on main? This is nullable on the main IIUC.

index.bs
```diff
@@ -356,6 +360,16 @@ interface SpeechRecognitionPhrase {
   </dd>
 </dl>
 
+<h4 id="speechrecoresult-attributes">SpeechRecognitionResult Attributes</h4>
+
+<dl>
+  <dt><dfn attribute for=SpeechRecognitionResult>speechStartTime</dfn> attribute</dt>
+  <dd>A {{double}} representing the start time of the speech segment corresponding to this recognition result, in seconds relative to the start of the audio stream consumed by the speech recognizer (where 0.0 seconds corresponds to the {{SpeechRecognition/onaudiostart}} event).</dd>
```
hoch: `onaudiostart` is th IDL attribute/handler, not the event name. I think we should reference the `audiostart` event itself.

index.bs
```diff
@@ -120,6 +120,8 @@ This does not preclude adding support for this as a future API enhancement, and
   <li>The user agent may also give the user a longer explanation the first time speech input is used, to let the user know what it is and how they can tune their privacy settings to disable speech recording if required.</li>
 
   <li>To mitigate the risk of fingerprinting, user agents MUST NOT personalize speech recognition when performing speech recognition on a {{MediaStreamTrack}}.</li>
+
+  <li>To mitigate micro-architectural timing attacks and hardware fingerprinting, user agents may reduce the resolution of {{SpeechRecognitionResult/speechStartTime}} and {{SpeechRecognitionResult/speechEndTime}} or introduce jitter, in accordance with the user agent's security and privacy policies (similar to [[HR-TIME-3]] and [[HTML]]).</li>
```
hoch: `may` should be capitalized to `MAY` (similar to MUST NOT in the item immediately preceding it) to clearly indicate RFC 2119 conformance.

index.bs
```diff
@@ -356,6 +360,16 @@ interface SpeechRecognitionPhrase {
   </dd>
 </dl>
 
+<h4 id="speechrecoresult-attributes">SpeechRecognitionResult Attributes</h4>
```
hoch: Is this the right location? What about L663?

`<h4 id="speechreco-result">SpeechRecognitionResult</h4>`

index.bs
alan33d: Done.

index.bs
alan33d: Done.

index.bs
alan33d: Done. Moved both attribute definitions into the <dl> under <h4 id="speechreco-result">SpeechRecognitionResult</h4> alongside length, item, and isFinal.

alan33d:
@hoch @evanbliu @padenot PTALA, thanks!

index.bs
```diff
@@ -120,6 +120,8 @@ This does not preclude adding support for this as a future API enhancement, and
   <li>The user agent may also give the user a longer explanation the first time speech input is used, to let the user know what it is and how they can tune their privacy settings to disable speech recording if required.</li>
 
   <li>To mitigate the risk of fingerprinting, user agents MUST NOT personalize speech recognition when performing speech recognition on a {{MediaStreamTrack}}.</li>
+
+  <li>To mitigate micro-architectural timing attacks and hardware fingerprinting, user agents MAY reduce the resolution of {{SpeechRecognitionResult/speechStartTime}} and {{SpeechRecognitionResult/speechEndTime}} or introduce jitter, in accordance with the user agent's security and privacy policies (similar to [[HR-TIME-3]] and [[HTML]]).</li>
```
padenot: This isn't doing much to mitigate anything, since it isn't related to a wallclock. Existing code are already running some form of VAD and/or full speech recognition engine (e.g. it's quite trivial in WASM) and can get the same numbers.

evanbliu (approved):

hoch (approved):
LGTM

index.bs
```diff
@@ -666,6 +670,12 @@ For example, some implementations may fire <a event for=SpeechRecognition>audioe
   <dt><dfn attribute for=SpeechRecognitionResult>isFinal</dfn> attribute</dt>
   <dd>The final boolean must be set to true if this is the final time the speech service will return this particular index value.
   If the value is false, then this represents an interim result that could still be changed.</dd>
+
+  <dt><dfn attribute for=SpeechRecognitionResult>speechStartTime</dfn> attribute</dt>
+  <dd>A {{double}} representing the start time of the speech segment corresponding to this recognition result, in seconds relative to the start of the audio stream consumed by the speech recognizer (where 0.0 seconds corresponds to the <a event for=SpeechRecognition>audiostart</a> event).</dd>
```
padenot: This isn't clear. Does `0.0` mean the time of the `"audiostart"` that already has a non-standard time, that can't really be implemented? E.g. what does it mean if the `MediaStreamTrack` comes from an `AudioContext`, or from an `HTMLMediaElement`, what time it is?

I agree that what you've described is maximally useful, but we might want to describe it in a non ambiguous way. I tried to do so but it went a bit out of hand so split it out in its own issue: https://github.com/WebAudio/web-speech-api/issues/213. Minimal to no changes are going to be required here, probably some editorial to refer to things more directly, but that's about it.

index.bs
alan33d: I don't have a strong opinion on this section so I removed it with a note. I agree and these are sample-count offsets on the session's audio timeline, not related to wallclock, so they can't be queried on demand for micro-architectural attacks.  And ack on page holding the MediaStreamTrack can already get sample-accurate boundaries from a WASM VAD.

index.bs
alan33d: Agreed, anchoring to audiostart ties the origin to when the engine finishes warming up rather than to the audio, and leaves AudioContext / HTMLMediaElement tracks ambiguous.

Both definitions now use the rule from #213: 0.0 is the position of the source audio at the moment the session began consuming it. This should keep it self-contained and forward-compatible if `SpeechRecognitionAudioEvent.time` lands.

alan33d:
@reillyeon @padenot PTALA, thanks!

reillyeon (approved):

alan33d:
@padenot PTALA and lmk if any other comments / concerns.  Thanks!

PR #195, opened 2026-07-15
Remove invisible characters from filename

reillyeon:
There were invisible characters at the end of speech-recognition-result-timestamps which were preserved when I added the .md extension. This change removes them.


reillyeon:
@hoch

padenot (approved):

PR #200, opened 2026-08-06
Run install() downloads in parallel instead of blocking the main thread.

### LINKED ISSUE #199: install() algorithm blocks main thread
[install()](https://webaudio.github.io/web-speech-api/#dom-speechrecognition-install) uses a novel promise pattern that blocks main thread:

Image

Step 6 uses [queue a task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task) (which queues another task on the main thread) when it probably meant [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel) (aka run in the background). The "When" statement implies waiting around, which stalls JS on main thread.

It would be clearer to run steps 5 and 6 [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel) and after the "when" happens, queue a task to resolve the promise (back on main).

padenot:
Step 6 queued a task on the relevant global object's task queue that then waited for the language pack downloads to complete, which stalls script on the main thread. Move the download initiation and the wait into an [=in parallel=] block, and queue a task only to resolve the promise.

Fixes #199

***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/pull/200.html" title="Last updated on Aug 6, 2026, 2:15 PM UTC (c377cb7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/200/8307ee1...c377cb7.html" title="Last updated on Aug 6, 2026, 2:15 PM UTC (c377cb7)">Diff</a>


padenot:
Added another commit that fixes a mistake in the prose, but it was clear what to do.

evanbliu (approved):

PR #203, opened 2026-08-12
Update audioStartTime and audioEndTime to be relative to start of aud…

alan33d:
We want to switch to  stream-Relative (0-based) impl. since:
- In Web Audio and Media APIs (HTMLMediaElement.currentTime, AudioContext.currentTime, WebCodecs VideoFrame.timestamp), media timelines are always 0-based offsets relative to stream start, not the time origin.
- Immune to inter-process jitter since SODA and audio capture run in a separate utility/browser process. Translating stream offsets to the renderer's timeOrigin relies on estimating when IPC AudioStarted() arrived, which introduces IPC latency jitter. Stream-relative offsets are not prune to this and aligns with the raw audio frames.
- If SpeechRecognition is used with a pre-recorded MediaStreamTrack, a 0-based stream offset reflects the actual position in the audio track regardless of when the webpage was loaded.
- Being relative to performance.timeOrigin doesn't make sense in general for the Web Speech API because it assumes that the audio source is live. Since a SpeechRecognizer can also be created for a prerecorded media stream I think the timestamps on the speech recognition events should be relative to the position in that media stream.

See https://crbug.com/542330168 for more details.

### DIFF
--- a/explainers/speech-recognition-result-timestamps.md
+++ b/explainers/speech-recognition-result-timestamps.md
@@ -39,7 +39,7 @@ recognition.interimResults = true;
 recognition.onresult = (event) => {
   const result = event.results[event.resultIndex];
 
-  if (result.audioEndTime !== null) {
+  if (result.audioEndTime !== null && result.audioEndTime !== undefined) {
     // Calculate on-device processing latency
     const processingLatencyMs = event.timeStamp - result.audioEndTime;
 
@@ -87,7 +87,7 @@ recognition.onaudiostart = (event) => {
 
 recognition.onresult = async (event) => {
   const result = event.results[event.resultIndex];
-  if (result.audioEndTime === null) return;
+  if (result.audioEndTime === null || result.audioEndTime === undefined) return;
 
   // 2. Convert stream offsets to document time origin coordinates
   const absoluteAudioStart = audioOriginTime + result.audioStartTime;
--- a/explainers/speech-recognition-result-timestamps.md
+++ b/explainers/speech-recognition-result-timestamps.md
@@ -54,6 +54,63 @@ recognition.onresult = (event) => {
 recognition.start();
 ```
 
+## Converting Stream Timestamps to Document Time Origin
+
+`audioStartTime` and `audioEndTime` are defined as media-local offsets in milliseconds relative to the start of the audio stream ($t = 0.0\text{ms}$).
+
+For real-time applications such as **live translation**, **subtitling overlays**, and **audio-visual sync**, developers often need to map these stream offsets to the document's global timeline (`DOMHighResTimeStamp` / `performance.now()`).
+
+### Pattern: Capturing the Audio Timeline Origin
+
+To convert stream-relative timestamps to document time coordinates:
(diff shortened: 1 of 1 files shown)


reillyeon (approved):

explainers/speech-recognition-result-timestamps.md
```diff
@@ -54,6 +54,62 @@ recognition.onresult = (event) => {
...
+let audioOriginTime = 0;
+
+// 1. Capture the audio stream's time origin on the document timeline
+recognition.onaudiostart = (event) => {
+  audioOriginTime = event.timeStamp;
+};
+
+recognition.onresult = async (event) => {
+  const result = event.results[event.resultIndex];
+  if (result.audioEndTime === null) return;
```
michaelwasserman: nit: use undefined instead of null

michaelwasserman (commented):
Drive by review: lgtm with a nit, fwiw. I implemented the suggested timestamp offsetting in my demo, and it seems to work well. https://michaelwasserman.github.io/web-speech-demos/speech_latency.html

explainers/speech-recognition-result-timestamps.md
alan33d: Because the Web IDL attribute is defined as nullable (DOMHighResTimeStamp?):

- undefined means the browser does not support the API or the feature flag is disabled.
- null means the browser supports the API, but timestamps are not available for this specific result.

So I think we want just `result.audioEndTime == null` which checks for both null and undefined?

reillyeon: You're both right. Given the WebIDL it will be `null` in the case where the browser doesn't provide a timestamp. However in the case where the browser hasn't been updated to support this API at all then it will be `undefined`.

IIRC the only reason we currently use `DOMHighResTimeStamp?` in the Chromium prototype is that we can't always provide timestamps due to an incomplete implementation. If we intend to fix that I'd recommend removing nullability from the WebIDL.

alan33d: Updated to `(result.audioEndTime !== null && result.audioEndTime !== undefined)` and `(result.audioEndTime === null && result.audioEndTime === undefined)` in case of any strict linting rules.

> IIRC the only reason we currently use DOMHighResTimeStamp? in the Chromium prototype is that we can't always provide timestamps due to an incomplete implementation. If we intend to fix that I'd recommend removing nullability from the WebIDL.

That's correct it can be null with incomplete impl. bugs like https://crbug.com/543076358.   But also for other reasons:

- Not All Speech Backends Support Timestamps: Chromium’s Web Speech API routes audio to different recognition backends: SODA vs. Cloud ASR vs. some other backend in Android? may not expose this timing data.
- If the attribute were non-nullable, an unsupported engine would have to return a placeholder like 0.0 or -1 (but 0.0ms is also a valid timestamp)
- Different on-device ML models have varying feature support e.g. SODA vs. TinyGemma.  Nullability allows some models to be used without breaking the API contract.

reillyeon (approved):

PR #206, opened 2026-09-01
Update speech recognition result timestamps explainer

alan33d:
Remove latency tracking uses cases since SpeechRecognitionEvent timestamps can be used for that purpose.  Instead focus on developer need for timeline association use cases in the updated explainer and alternatives considered.


alan33d (commented):
This will be merged back to the main Webspeech repo

evanbliu (approved):

reillyeon (approved):

PR #212, opened 2026-09-08
Make SpeechGrammarList.item() nullable.

padenot:
The return type was non-nullable, so an index greater than or equal to length had no valid return value. Chromium declares it non-nullable too, but its implementation returns null for such an index, and other platform collections with an item() getter (NodeList, HTMLCollection, CSSRuleList, FileList) all declare it nullable, so specify null and match what ships.

I'm sending a Gecko patch that includes a WPT for this.

***
<a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/pull/212.html" title="Last updated on Sep 8, 2026, 4:48 PM UTC (3f9e1f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebAudio/web-speech-api/212/bcdfb7f...3f9e1f4.html" title="Last updated on Sep 8, 2026, 4:48 PM UTC (3f9e1f4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -256,7 +256,7 @@ interface SpeechRecognitionAlternative {
 [SecureContext, Exposed=Window]
 interface SpeechRecognitionResult {
     readonly attribute unsigned long length;
-    getter SpeechRecognitionAlternative item(unsigned long index);
+    getter SpeechRecognitionAlternative? item(unsigned long index);
     readonly attribute boolean isFinal;
 };
 
@@ -264,7 +264,7 @@ interface SpeechRecognitionResult {
 [SecureContext, Exposed=Window]
 interface SpeechRecognitionResultList {
     readonly attribute unsigned long length;
-    getter SpeechRecognitionResult item(unsigned long index);
+    getter SpeechRecognitionResult? item(unsigned long index);
 };
 
 // A full response, which could be interim or final, part of a continuous response or not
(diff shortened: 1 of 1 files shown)


padenot:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/speech/speech_grammar_list.cc;l=39-41
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/speech/speech_grammar_list.idl;drc=fc7461989778f18435f25f472eac3bde815afa17;l=36

is what chrome do for completeness, this is equivalent I think.

index.bs
```diff
@@ -292,7 +292,7 @@ interface SpeechGrammar {
 interface SpeechGrammarList {
     constructor();
     readonly attribute unsigned long length;
-    getter SpeechGrammar item(unsigned long index);
+    getter SpeechGrammar? item(unsigned long index);
```
hoch, line 295: Perhaps we can follow up on these as well?

SpeechRecognitionResult
SpeechRecognitionResultList

It looks like they are all nullable too.

hoch (approved):
LGTM

index.bs
padenot: yeah good call, added.

